PR #68, opened 2020-05-20
fix: use right x-version-update in POM files

anguillanneuf:
Use `<!-- {x-version-update:google-cloud-pubsublite:released} -->` in `samples/install-without-bom`, and `samples/snippets` (for now, to be removed later).

Use `<!-- {x-version-update:google-cloud-pubsublite:current} -->` in `samples/snapshot`. 

---

Kokoro - Test: Java 11 and Kokoro - Test: Java 8 are flaky. Kokoro - CI seems stuck Other libraries have migrated to using GitHub actions for these.

### DIFF
--- a/samples/snippets/pom.xml
+++ b/samples/snippets/pom.xml
@@ -64,7 +64,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-pubsublite</artifactId>
-      <version>0.1.4</version><!-- {x-version-update:google-cloud-pubsublite:current} -->
+      <version>0.1.4</version><!-- {x-version-update:google-cloud-pubsublite:released} -->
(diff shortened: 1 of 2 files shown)


samples/snippets/pom.xml
```diff
@@ -64,7 +64,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-pubsublite</artifactId>
-      <version>0.1.4</version><!-- {x-version-update:google-cloud-pubsublite:current} -->
+      <version>0.1.4</version><!-- {x-version-update:google-cloud-pubsublite:released} -->
```
anguillanneuf, line 67: Remove after Pub/Sub Lite works with libraries-bom.

Issue #7, opened 2020-04-17
AdminClient fails shut down / warnings at shutdown


anguillanneuf:
After calling `.get()` on ApiFutures returned by AdminClient operations such as `getTopicPartitionCount` and `createTopic`, the program should exit, but it doesn't. Those operations succeed though.

PR #72, opened 2020-05-21
fix: Implement ChannelCache, which aggregates stubs to a single channel, and properly cleans them up on teardown.

dpcollins-google:


anguillanneuf:
Fixes #7

PR #77, opened 2020-05-21
chore: regenerate README from templates

chingor13:
Dependent on adding the `min_java_version` in https://github.com/googleapis/synthtool/pull/577


.repo-metadata.json
```diff
@@ -1,14 +1,16 @@
 {
   "name": "pubsublite",
   "name_pretty": "Cloud Pub/Sub Lite",
-  "product_documentation": "n/a",
-  "api_description": "n/a",
+  "product_documentation": "https://cloud.google.com/pubsub",
```
anguillanneuf, line 4: This should be "https://cloud.google.com/pubsub/lite". It will go public next week.

README.md
```diff
@@ -330,35 +330,101 @@ try {
...
+## CI Status
+
+Java Version | Status
+------------ | ------
+Java 8 | [![Kokoro CI][kokoro-badge-image-2]][kokoro-badge-link-2]
+Java 8 OSX | [![Kokoro CI][kokoro-badge-image-3]][kokoro-badge-link-3]
+Java 8 Windows | [![Kokoro CI][kokoro-badge-image-4]][kokoro-badge-link-4]
+Java 11 | [![Kokoro CI][kokoro-badge-image-5]][kokoro-badge-link-5]
+
+[product-docs]: https://cloud.google.com/pubsub
```
anguillanneuf, line 404: This should be https://cloud.google.com/pubsub/lite

samples/pom.xml
```diff
@@ -45,7 +45,6 @@
   </properties>
 
   <modules>
-    <module>install-without-bom</module>
     <module>snapshot</module>
     <module>snippets</module>
   </modules>
```
anguillanneuf, line 50: Why is "install-without-bom" getting removed?

samples/snippets/pom.xml
```diff
@@ -40,31 +40,18 @@
...
-      <dependency>
-        <groupId>com.google.cloud</groupId>
-        <artifactId>libraries-bom</artifactId>
-        <version>5.3.0</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
-
```
anguillanneuf, line 58: Is `libraries-bom` not recommended for this library? Why?

samples/snippets/pom.xml
```diff
@@ -40,31 +40,18 @@
...
-      <artifactId>google-cloud-pubsub</artifactId>
+      <artifactId>google-cloud-pubsublite</artifactId>
+      <version>0.1.4</version>
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
-      <artifactId>google-cloud-pubsublite</artifactId>
-      <version>0.1.4</version><!-- {x-version-update:google-cloud-pubsublite:released} -->
+      <artifactId>google-cloud-pubsub</artifactId>
+      <version>1.105.0</version>
```
anguillanneuf, line 54: I assume there is not need for this tag `<!-- {x-version-update:google-cloud-pubsublite:released} -->`?

.readme-partials.yaml
```diff
@@ -0,0 +1,259 @@
...
+  Topic topic =
+      Topic.newBuilder()
+        .setPartitionConfig(
+          PartitionConfig.newBuilder()
+            .setScale(1) // Set publishing throughput to 1*4 MiB per sec. This must be 1-4.
+            .setCount(2)) // The number of partitions.
+        .setRetentionConfig(
+          RetentionConfig.newBuilder()
+            .setPeriod(Durations.fromDays(7))
+            .setPerPartitionBytes(100000000000L)) // 100 GiB. This must be 30 GiB-10 TiB.
```
anguillanneuf, line 33: This should be 100L * 1024L * 1024L. Please reference the updated samples in 

https://github.com/googleapis/java-pubsublite/blob/caa96a86feea501daa2390a1a3ca92142b8f14b6/samples/snippets/src/main/java/pubsublite/CreateTopicExample.java#L83

.readme-partials.yaml
```diff
@@ -0,0 +1,259 @@
...
+  Then, to publish messages asynchronously, use the following code:
+
+  ```java
+  public class PublisherExample {
+    private static final int MESSAGE_COUNT = 10;
+
+    // Load the project number from a commandline flag.
+    private static final long PROJECT_NUMBER = 123L;
+    // Load the zone from a commandline flag.
+    private static final String ZONE = "us-central1-b";
```
anguillanneuf, line 66: ZONE_ID to avoid confusion with Cloud Zone.

samples/pom.xml
chingor13: We are not adding any pre-1.0.0 libraries to the google-cloud-bom and thus the libraries-bom. Therefore, the snippets module now contains the direct installation method.

Once this library hits 1.0 and we include it in the home, then we'll bring back the libraries-bom samples and tests. The presence of the libraries bom snippet will cause those instructions to appear in the README

samples/snippets/pom.xml
chingor13: The released version will be updated by renovate bot after it's available in Maven Central. Otherwise, these tests will fail immediately after a release because the released version is not found

.readme-partials.yaml
chingor13: Sounds good. I just cut this content from the existing README. Will adjust accordingly.

samples/pom.xml
anguillanneuf: FYI: @soyrice 

We need to update the library installation section accordingly.

.repo-metadata.json
soyrice: The full link to the docs will be be https://cloud.google.com/pubsub/lite/docs

https://cloud.google.com/pubsub links to the marketing page, and Pub/Sub and Pub/Sub Lite will share a marketing page, so https://cloud.google.com/pubsub/lite will redirect to https://cloud.google.com/pubsub

samples/snippets/pom.xml
anguillanneuf: I see now. Thanks!

anguillanneuf:
Just the docs link that needs an update. LGTM.

PR #81, opened 2020-05-26
fix: Return an error on publish if the publisher's state is not RUNNING

tmdiep:

Fixes internal issue b/156525288.


tmdiep:
Not sure if this is the best solution.
I didn't call onPermanentError() because the publisher may already be in a failed or terminated state.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/WrappingPublisher.java
```diff
@@ -55,6 +57,15 @@ protected void handlePermanentError(StatusException error) {}
   // Publisher implementation.
   @Override
   public ApiFuture<String> publish(PubsubMessage message) {
+    ApiService.State currentState = state();
```
dpcollins-google, line 60: I don't think this is the right layer to enforce this at. The WrappingPublisher is just a thin wrapper around the internal publisher. Perhaps add this in the wire publisher implementation instead?

tmdiep: Done. Moved the check to PublisherImpl.

dpcollins-google:
> Not sure if this is the best solution.
> I didn't call onPermanentError() because the publisher may already be in a failed or terminated state.

This should be okay. ProxyService will handle this gracefully, please add onPermanentError.

tmdiep:
> > I didn't call onPermanentError() because the publisher may already be in a failed or terminated state.
> 
> This should be okay. ProxyService will handle this gracefully, please add onPermanentError.

Done. PTAL at revised implementation in PublisherImpl.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PublisherImpl.java
```diff
@@ -209,12 +210,12 @@ private void processBatch(Collection<UnbatchedMessage> batch) throws StatusExcep
...
-        return ApiFutures.immediateFailedFuture(
-            Status.FAILED_PRECONDITION
-                .withDescription("Published after the stream shut down.")
-                .asException());
-      }
+      ApiService.State currentState = state();
+      checkState(
+          currentState == ApiService.State.RUNNING,
+          String.format("Cannot publish when Publisher state is %s.", currentState.name()));
+      // This is an assertion. If the service is running, it should not be shutdown.
```
dpcollins-google, line 217: Please remove this comment. This is redundant.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PublisherImpl.java
```diff
@@ -236,6 +237,13 @@ void flushToStream() {
     }
   }
 
+  @VisibleForTesting
```
dpcollins-google, line 240: Please remove this method. You can check the permanent error handlerr to determine if a permanent error occurred.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PublisherImpl.java
```diff
@@ -209,12 +210,12 @@ private void processBatch(Collection<UnbatchedMessage> batch) throws StatusExcep
...
     try (CloseableMonitor.Hold h = monitor.enter()) {
-      if (shutdown) {
-        return ApiFutures.immediateFailedFuture(
-            Status.FAILED_PRECONDITION
-                .withDescription("Published after the stream shut down.")
-                .asException());
-      }
+      ApiService.State currentState = state();
+      checkState(
+          currentState == ApiService.State.RUNNING,
```
dpcollins-google, line 215: checkState(state() == ApiService.State.RUNNING, ...);

tmdiep: The state is also used for the error message. I saw that state() does some enum mappings that we could avoid doing twice:
https://github.com/googleapis/api-common-java/blob/master/src/main/java/com/google/api/core/AbstractApiService.java#L125

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PublisherImpl.java
tmdiep: Yeah, that was the first thing I tried, but the listeners don't get called. I traced through and think it's because [AbstractService::notifyFailed](https://github.com/google/guava/blob/master/android/guava/src/com/google/common/util/concurrent/AbstractService.java#L451) just throws an IllegalStateException when the state is New or Terminated, and doesn't transition to Failed. Since there's no state transition, the listeners aren't called.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PublisherImpl.java
tmdiep: Done

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PublisherImpl.java
dpcollins-google: SGTM.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PublisherImpl.java
dpcollins-google: Put it into STARTING or STOPPING by making the startAsync or stopAsync method of the RetryingConnection block on a CountdownLatch. Then try to publish to trigger this case, and then let the method proceed.

tmdiep: Thanks, CountDownLatch is useful. I looked into your suggestion and noticed that all of our doStarts are synchronous. The only scenario that really triggers the original bug (b/156525288) is when the user publishes before calling startAsync, and so left the test to check NEW.

I removed the isShutdown() and just verified that the service has failed permanently by checking that an IllegalStateException is thrown if the user then tries to start the service. This occurs if they try to start PublisherImpl or any proxy such as WrappingPublisher.

PR #87, opened 2020-05-26
fix: keep track of internal seek

hannahrogers-google:
If a seek is inFlight and a stream is retried, retrying the inFlight seek.

If an internal seek is in flight, and a client initiates a seek, block until the internal seek completes.

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
@@ -143,6 +143,7 @@ public class SubscriberImpl extends ProxyService
       checkState(!inFlightSeek.isPresent(), "Seeked while seek is already in flight.");
       SettableApiFuture<Offset> future = SettableApiFuture.create();
       inFlightSeek = Optional.of(new InFlightSeek(request, future));
+      tokenCounter.onClientSeek();
       connection.modifyConnection(
           connectedSubscriber ->
               connectedSubscriber.ifPresent(subscriber -> subscriber.seek(request)));
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
@@ -34,6 +34,7 @@ import com.google.cloud.pubsublite.proto.SubscribeRequest;
 import com.google.cloud.pubsublite.proto.SubscriberServiceGrpc.SubscriberServiceStub;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.Monitor;
 import io.grpc.Status;
 import io.grpc.StatusException;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
```diff
@@ -115,13 +129,20 @@ protected void stop() {
 
   @Override
   public ApiFuture<Offset> seek(SeekRequest request) {
```
manuelmenzella-google, line 131: Should we reset the tokenCounter on seeks()?

dpcollins-google: Yes, almost certainly.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
```diff
@@ -115,13 +129,20 @@ protected void stop() {
...
-    try (CloseableMonitor.Hold h = monitor.enter()) {
+    try (CloseableMonitor.Hold h =
+        monitor.enterWhenUninterruptibly(
+            new Monitor.Guard(monitor.monitor) {
+              @Override
+              public boolean isSatisfied() {
+                return !internalSeekInFlight;
```
dpcollins-google, line 137: This should alternately check shutdown, or it can stall forever on a permanent failure.

hannahrogers-google: Good catch - thanks! Fixed.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
hannahrogers-google: Would we want to do this when a seek request is sent, or when a seek response is received?  If a client initiates a seek, and then immediately calls allowFlow before receiving a seek response, what is the expectation here?

manuelmenzella-google: We should do it as soon as the client initiates the seek, IMO. If you think about the stream, messages are sent in order, so the flow-control should be processed after the seek.

hannahrogers-google: sounds good thanks!  I have added a tokenCounter reset on seek

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
```diff
@@ -115,13 +129,21 @@ protected void stop() {
 
   @Override
   public ApiFuture<Offset> seek(SeekRequest request) {
-    try (CloseableMonitor.Hold h = monitor.enter()) {
+    try (CloseableMonitor.Hold h =
+        monitor.enterWhenUninterruptibly(
+            new Monitor.Guard(monitor.monitor) {
+              @Override
+              public boolean isSatisfied() {
+                return !internalSeekInFlight || shutdown;
```
manuelmenzella-google, line 137: If shutdown == true, we'll checkState-fail right below this. I think we should return an immediately-failed Future.

hannahrogers-google: I think this should already happen.  The catch block will catch the checkState failure and will return an immediateFailedFuture.

PR #92, opened 2020-06-03
Implements batching of flow control requests

tmdiep:

Fixes internal issue b/157362915.

I still need to write the tests, but this is a preview of both of the proposed approaches. I think it's not too complex.

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
@@ -29,10 +29,10 @@ class FlowControlBatcher {
   private static double EXPEDITE_BATCH_REQUEST_RATIO = 0.5;
 
   // The current amount of outstanding byte and message flow control tokens.
-  private TokenCounter clientTokens = new TokenCounter();
+  private final TokenCounter clientTokens = new TokenCounter();
 
   // The pending aggregate flow control request that needs to be sent to the stream.
-  private TokenCounter pendingTokens = new TokenCounter();
+  private final TokenCounter pendingTokens = new TokenCounter();
 
   void onClientFlowRequest(FlowControlRequest request) throws StatusException {
     clientTokens.add(request.getAllowedBytes(), request.getAllowedMessages());
@@ -49,13 +49,13 @@ class FlowControlBatcher {
     pendingTokens.reset();
   }
 
-  // The caller must send the FlowControlRequest to the stream.
-  Optional<FlowControlRequest> releaseRequestForRestart() {
+  // The caller must send the FlowControlRequest to the stream, as pending tokens are reset.
+  Optional<FlowControlRequest> requestForRestart() {
     pendingTokens.reset();
     return clientTokens.toFlowControlRequest();
   }
 
-  // The caller must send the FlowControlRequest to the stream.
+  // The caller must send the FlowControlRequest to the stream, as pending tokens are reset.
   Optional<FlowControlRequest> releasePendingRequest() {
     Optional<FlowControlRequest> request = pendingTokens.toFlowControlRequest();
     pendingTokens.reset();
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
@@ -29,48 +29,46 @@ class FlowControlBatcher {
   private static double EXPEDITE_BATCH_REQUEST_RATIO = 0.5;
 
   // The current amount of outstanding byte and message flow control tokens.
-  private TokenCounter currentOutstandingTokens = new TokenCounter();
+  private TokenCounter clientTokens = new TokenCounter();
 
   // The pending aggregate flow control request that needs to be sent to the stream.
-  private TokenCounter pendingFlowRequest = new TokenCounter();
+  private TokenCounter pendingTokens = new TokenCounter();
 
   void onClientFlowRequest(FlowControlRequest request) throws StatusException {
-    currentOutstandingTokens.add(request.getAllowedBytes(), request.getAllowedMessages());
-    pendingFlowRequest.add(request.getAllowedBytes(), request.getAllowedMessages());
+    clientTokens.add(request.getAllowedBytes(), request.getAllowedMessages());
+    pendingTokens.add(request.getAllowedBytes(), request.getAllowedMessages());
   }
 
   void onMessages(Collection<SequencedMessage> received) throws StatusException {
     long byteSize = received.stream().mapToLong(SequencedMessage::byteSize).sum();
-    currentOutstandingTokens.sub(byteSize, received.size());
+    clientTokens.sub(byteSize, received.size());
   }
 
   void onClientSeek() {
-    currentOutstandingTokens.reset();
-    pendingFlowRequest.reset();
+    clientTokens.reset();
+    pendingTokens.reset();
   }
 
-  // Must be called after a FlowControlRequest has been sent to the stream.
-  void onFlowRequestDispatched() {
-    pendingFlowRequest.reset();
+  // The caller must send the FlowControlRequest to the stream.
+  Optional<FlowControlRequest> releaseRequestForRestart() {
+    pendingTokens.reset();
+    return clientTokens.toFlowControlRequest();
   }
 
-  Optional<FlowControlRequest> requestForRestart() {
-    return currentOutstandingTokens.toFlowControlRequest();
-  }
-
-  Optional<FlowControlRequest> pendingBatchRequest() {
-    return pendingFlowRequest.toFlowControlRequest();
+  // The caller must send the FlowControlRequest to the stream.
+  Optional<FlowControlRequest> releasePendingRequest() {
+    Optional<FlowControlRequest> request = pendingTokens.toFlowControlRequest();
+    pendingTokens.reset();
+    return request;
   }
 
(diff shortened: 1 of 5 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
```diff
@@ -252,4 +278,28 @@ private Status onSeekResponse(Offset seekOffset) {
...
     }
   }
+
+  @VisibleForTesting
+  void processBatchFlowRequest() {
+    try (CloseableMonitor.Hold h = monitor.enter()) {
+      if (shutdown) return;
+      connection.modifyConnection(
+          connectedSubscriber ->
+              connectedSubscriber.ifPresent(subscriber -> flushBatchFlowRequest(subscriber)));
```
manuelmenzella-google, line 288: Won't this deadlock? flushBatchFlowRequest also tries to grab the mutex. It would be interesting to exercise this on the text, if possible.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
```diff
@@ -0,0 +1,76 @@
...
+    currentOutstandingTokens.reset();
+    pendingFlowRequest.reset();
+  }
+
+  // Must be called after a FlowControlRequest has been sent to the stream.
+  void onFlowRequestDispatched() {
+    pendingFlowRequest.reset();
+  }
+
+  Optional<FlowControlRequest> requestForRestart() {
```
manuelmenzella-google, line 57: Just curious: why are these optionals?

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
```diff
@@ -0,0 +1,76 @@
...
+class FlowControlBatcher {
+  // If the pending flow request exceeds this ratio of the current outstanding tokens, the pending
+  // flow request should be flushed to the stream ASAP.
+  private static double EXPEDITE_BATCH_REQUEST_RATIO = 0.5;
+
+  // The current amount of outstanding byte and message flow control tokens.
+  private TokenCounter currentOutstandingTokens = new TokenCounter();
+
+  // The pending aggregate flow control request that needs to be sent to the stream.
+  private TokenCounter pendingFlowRequest = new TokenCounter();
```
manuelmenzella-google, line 35: Nit: the naming is a bit inconsistent here; how about clientTokens and pendingTokens?

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
```diff
@@ -0,0 +1,76 @@
+/*
```
manuelmenzella-google, line 1: Can we have a test for this file?

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
dpcollins-google: java locks are reentrant. But we should exercise this in the test, if it is not.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
dpcollins-google: There can be no bytes or messages, which we wouldn't want to send a request for.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
tmdiep: Yep, the [Monitor](https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/Monitor.java#L35) that we use is reentrant. Existing tests do cover this and they pass.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
```diff
@@ -0,0 +1,76 @@
...
+  // Must be called after a FlowControlRequest has been sent to the stream.
+  void onFlowRequestDispatched() {
+    pendingFlowRequest.reset();
+  }
+
+  Optional<FlowControlRequest> requestForRestart() {
+    return currentOutstandingTokens.toFlowControlRequest();
+  }
+
+  Optional<FlowControlRequest> pendingBatchRequest() {
```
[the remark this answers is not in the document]
tmdiep: Hmm... I considered that and was concerned that the request may fail and we end up dropping a batch flow request. Is there any chance that ConnectedSubscriber::allowFlow fails transiently? Or if it does, would the entire connection be reinitialized?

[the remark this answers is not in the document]
tmdiep: I'll assume the answer is yes, the [stream will be terminated](https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/StreamObserver.html#onNext-V-) and the connection reinitialized, and will change to releasePendingRequest and also releaseRequestForRestart, as onFlowRequestDispatched needs to be called for both.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
tmdiep: Done.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
```diff
@@ -0,0 +1,76 @@
...
+package com.google.cloud.pubsublite.internal.wire;
+
+import com.google.cloud.pubsublite.SequencedMessage;
+import com.google.cloud.pubsublite.proto.FlowControlRequest;
+import io.grpc.StatusException;
+import java.util.Collection;
+import java.util.Optional;
+
+// A FlowControlBatcher manages batching of FlowControlRequests.
+class FlowControlBatcher {
```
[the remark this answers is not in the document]
tmdiep: I was able to validate everything I wanted to by verifying the calls to mockConnectedSubscriber::allowFlow. Also added more comprehensive tests in FlowControlBatcherTest.java, so I think we don't need the interface, at least for now anyway.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
tmdiep: Done.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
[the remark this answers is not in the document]
tmdiep: Done.

tmdiep:
Thanks for the review. Comments addressed and tests completed now. PTAL.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
```diff
@@ -0,0 +1,74 @@
...
+    clientTokens.sub(byteSize, received.size());
+  }
+
+  void onClientSeek() {
+    clientTokens.reset();
+    pendingTokens.reset();
+  }
+
+  // The caller must send the FlowControlRequest to the stream.
+  Optional<FlowControlRequest> releaseRequestForRestart() {
```
dpcollins-google, line 53: nit. there's no ownership transfer here, consider changing the name back?

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/TokenCounter.java
```diff
@@ -19,43 +19,50 @@
...
+    checkState(deltaMessages <= messages, "Received more messages than were requested.");
 
-    bytes -= byteSize;
-    messages -= received.size();
+    bytes -= deltaBytes;
+    messages -= deltaMessages;
   }
 
-  void onClientSeek() {
+  void reset() {
```
dpcollins-google, line 52: nit. I wonder whether this needs to be a method, or if you can just do var = new TokenCounter()

tmdiep: Either is fine. Slight preference for avoiding the creation of new instances, since the reset is simple.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java
tmdiep: Done. I was trying to communicate "caller must send to stream", but I see your point. Reverted the name and just updated the comment instead.

tmdiep:
Thanks again for the review. I just addressed the nits in this patch.

PR #95, opened 2020-06-03
fix: use appropriate x-version-update tag

anguillanneuf:
Add `<!-- {x-version-update-start} -->` and `<!-- {x-version-update-end} -->` in places where they were missed. 

This is probably the reason versions weren't updated in samples POM and the reason autosynth couldn't update README.

---
Renovate bot is supposed to update samples POM. Why is it not?

### DIFF
--- a/samples/snapshot/pom.xml
+++ b/samples/snapshot/pom.xml
@@ -48,11 +48,13 @@
       <version>0.1.7-SNAPSHOT</version>
     </dependency>
     <!-- {x-version-update-end} -->
+    <!-- {x-version-update-start:google-cloud-pubsublite:released} -->
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-pubsub</artifactId>
       <version>1.105.0</version>
     </dependency>
+    <!-- {x-version-update-end} -->
 
     <!-- A logging dependency used by the underlying library  -->
     <dependency>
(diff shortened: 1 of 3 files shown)


anguillanneuf:
@chingor13 Is it okay to close chore: release 0.1.7 #93 to cause it to rerun with the changes made in this PR?

anguillanneuf:
@chingor13 That makes sense. In that case, how do I get renovate bot to work appropriately? 

That last commit may still be awry.

samples/snapshot/pom.xml
```diff
@@ -48,11 +48,13 @@
       <version>0.1.7-SNAPSHOT</version>
     </dependency>
     <!-- {x-version-update-end} -->
+    <!-- {x-version-update-start:google-cloud-pubsub:released} -->
```
dpcollins-google, line 51: This refers to cloud pubsub not pubsublite. Is this intentional?

anguillanneuf: This tag works by picking up version numbers from `versions.txt`. 
I'm guessing this is not needed because renovate bot is supposed to update deps in POMs. I may need to close this PR, but I will wait for an answer from Jeff.

anguillanneuf:
@chingor and I talked offline. Not sure why Renovate bot isn't updating library versions in samples POM. I will update them manually for the time being.

PR #122, opened 2020-06-09
[fix] Disable the clirr plugin causing a build failure before GA.

dpcollins-google:
We will have to make breaking changes before GA. This plugin would require adding an explicit xml annotation for all of those that happen, which is unreasonable in the short term, given that removing a beam class triggered 6 distinct errors that would need to be silenced.

8001: com.google.cloud.pubsublite.beam.DeduplicationFnOptions: Class com.google.cloud.pubsublite.beam.DeduplicationFnOptions removed
8001: com.google.cloud.pubsublite.beam.DeduplicationFnOptions$Builder: Class com.google.cloud.pubsublite.beam.DeduplicationFnOptions$Builder removed
7013: com.google.cloud.pubsublite.beam.UuidDeduplicationOptions: Abstract method 'public org.apache.beam.sdk.transforms.Deduplicate$KeyedValues deduplicate()' has been added
7002: com.google.cloud.pubsublite.beam.UuidDeduplicationOptions: Method 'public com.google.cloud.pubsublite.beam.DeduplicationFnOptions deduplicationFnOptions()' has been removed
7013: com.google.cloud.pubsublite.beam.UuidDeduplicationOptions$Builder: Abstract method 'public com.google.cloud.pubsublite.beam.UuidDeduplicationOptions$Builder setDeduplicate(org.apache.beam.sdk.transforms.Deduplicate$KeyedValues)' has been added
7002: com.google.cloud.pubsublite.beam.UuidDeduplicationOptions$Builder: Method 'public com.google.cloud.pubsublite.beam.UuidDeduplicationOptions$Builder setDeduplicationFnOptions(com.google.cloud.pubsublite.beam.DeduplicationFnOptions)' has been removed


chingor13:
You can alternatively explicitly allow certain breaking change: https://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html

PR #135, opened 2020-06-11
feat: Implement AssigningSubscriber which uses the PartitionAssignmentService to determine which partitions to connect to.

dpcollins-google:
Diffbased on #134

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AssigningSubscriber.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AssigningSubscriber.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.pubsublite.cloudpubsub.internal;
+
+import static com.google.cloud.pubsublite.internal.ExtractStatus.toCanonical;
+import static com.google.cloud.pubsublite.internal.Preconditions.checkState;
+
+import com.google.cloud.pubsublite.Partition;
+import com.google.cloud.pubsublite.cloudpubsub.Subscriber;
+import com.google.cloud.pubsublite.internal.CloseableMonitor;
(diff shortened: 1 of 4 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AssigningSubscriber.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  private void handleAssignment(Set<Partition> assignment) {
+    try {
+      try (CloseableMonitor.Hold h = monitor.enter()) {
+        if (shutdown) return;
+        Set<Partition> livePartitions = ImmutableSet.copyOf(liveSubscriberMap.keySet());
+        for (Partition partition : livePartitions) {
+          if (!assignment.contains(partition)) {
+            stopSubscriber(liveSubscriberMap.remove(partition));
```
manuelmenzella-google, line 85: I'm a bit worried about acking the assignment after calling stopAsync, and not waiting for the Subscriber to actually stop. I understand that we now tear down the stream immediately, but I don't think we should rely on that externally, since it's an implementation detail.

Have you considered acking assignments asynchronously? It would be a bit more complex, but I think it's the right thing to do.

dpcollins-google: I don't think it is the right thing to do. This isn't relying on the fact that we stop synchronously for correctness: All that matters is after stopAsync is called, we won't retry on the stream, so an eventual state in which stopAsync() set a "shutdown" boolean but didn't wait for the stream to stop is still correct.

Acking asynchronously adds some complexity that I don't think is needed since the work is bounded. WDYT?

manuelmenzella-google: Yeah, I think that's fair. "All that matters is after stopAsync is called, we won't retry on the stream"; that still sounds like implementation details, but I agree that the alternative is a lot more complex.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AssigningSubscriber.java
```diff
@@ -0,0 +1,123 @@
...
+    subscriber.addListener(
+        new Listener() {
+          @Override
+          public void failed(State from, Throwable failure) {
+            onPermanentError(toCanonical(failure));
+          }
+
+          @Override
+          public void terminated(State from) {
+            stoppingSubscribers.remove(subscriber);
```
manuelmenzella-google, line 110: I'm a bit worried about having multiple "stopping subscribers" for the same partition, on assignment churn. WDYT?

dpcollins-google: I'm not particularly, although this code handles that case since its checking object identity not partition number. The subsequent subscribers will force-evict older ones, so will just speed up them shutting down by failing their streams with an error, so it should be okay.

PR #153, opened 2020-06-29
test: Add VPC-SC Negative Integration Tests

ORabbit:
Add integration tests to verify successful rejection of service when a project is being used outside its VPC perimeter.

These tests are **NOT** run automatically and require the env variable `GOOGLE_CLOUD_TESTS_IN_VPCSC=true` to be set. Additionally, the project to run the tests against can be set with env variable `GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT_NUMBER`.

A simple way to run these tests is as follows:
```bash
export GOOGLE_APPLICATION_CREDENTIALS=/path/to/service/account.json
export GOOGLE_CLOUD_TESTS_IN_VPCSC=true
export GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT_NUMBER=<project-number>
mvn clean -Dtest=ITVPCNegativeTest -DfailIfNoTests=false test
```

Internal b/159125512.

### DIFF
--- a/google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
+++ b/google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
@@ -155,6 +155,7 @@ public class ITVPCNegativeTest {
                 DeliveryConfig.newBuilder()
                     .setDeliveryRequirement(DeliveryRequirement.DELIVER_AFTER_STORED))
             .setName(subscriptionPath.value())
+            .setTopic(topicPath.value())
             .build();
 
     // Instantiate an AdminClient to test with.
@@ -163,7 +164,9 @@ public class ITVPCNegativeTest {
   }
 
   @After
-  public void tearDown() throws Exception {}
+  public void tearDown() throws Exception {
+    adminClient.close();
+  }
 
   private void checkExceptionForVPCError(StatusRuntimeException e) {
     assertEquals(Status.Code.PERMISSION_DENIED, e.getStatus().getCode());
--- a/google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
+++ b/google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.pubsublite.it;
 
+import static com.google.cloud.pubsublite.internal.ExtractStatus.toCanonical;
 import static com.google.common.truth.Truth.assertThat;
 import static junit.framework.TestCase.assertNotNull;
 import static org.junit.Assert.assertEquals;
@@ -179,14 +180,13 @@ public class ITVPCNegativeTest {
     } catch (InterruptedException e) {
       fail("Expected PERMISSION_DENIED StatusRuntimeException but got: " + e.toString());
     } catch (ExecutionException e) {
-      Throwable thrown = e.getCause();
-      checkExceptionForVPCError((StatusRuntimeException) thrown);
+      checkExceptionForVPCError(toCanonical(e.getCause()));
       return;
     }
 
     // If we successfully create a topic we need to clean it up.
     try {
-      adminClient.deleteSubscription(subscriptionPath).get();
+      adminClient.deleteTopic(topicPath).get();
       fail("Expected PERMISSION_DENIED StatusRuntimeException during createTopic");
     } catch (Exception e) {
       fail(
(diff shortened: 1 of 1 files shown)


google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
```diff
@@ -0,0 +1,391 @@
...
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.Timeout;
+
+/* Integration tests for VPC-SC */
+public class ITVPCNegativeTest {
+
+  private static final boolean IS_VPCSC_TEST =
```
dpcollins-google, line 73: sorry, when would this NOT be true?

google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
```diff
@@ -0,0 +1,391 @@
...
+    } catch (ExecutionException e) {
+      Throwable thrown = e.getCause();
+      checkExceptionForVPCError((StatusRuntimeException) thrown);
+      return;
+    }
+
+    // If we successfully create a topic we need to clean it up.
+    try {
+      adminClient.deleteSubscription(subscriptionPath).get();
+      fail("Expected PERMISSION_DENIED StatusRuntimeException during createTopic");
```
dpcollins-google, line 190: deleteTopic

google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
```diff
@@ -0,0 +1,391 @@
...
+      fail("Expected PERMISSION_DENIED StatusRuntimeException but got: " + e.toString());
+    } catch (ExecutionException e) {
+      Throwable thrown = e.getCause();
+      checkExceptionForVPCError((StatusRuntimeException) thrown);
+      return;
+    }
+
+    // If we successfully create a topic we need to clean it up.
+    try {
+      adminClient.deleteSubscription(subscriptionPath).get();
```
dpcollins-google, line 189: this call (once fixed) should be inside the 'try' block above

google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
```diff
@@ -0,0 +1,391 @@
...
+      fail("Expected PERMISSION_DENIED StatusRuntimeException but got: " + e.toString());
+    } catch (ExecutionException e) {
+      Throwable thrown = e.getCause();
+      checkExceptionForVPCError((StatusRuntimeException) thrown);
+      return;
+    }
+
+    // If we successfully create a subscription we need to clean it up.
+    try {
+      adminClient.deleteSubscription(subscriptionPath).get();
```
dpcollins-google, line 278: see comment above

google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
ORabbit: In most situation this would NOT be true. These integration tests should only be run on projects that have VPC-SC enabled for their GCP organization + underlying project. Since the average user will not be using VPC-SC these tests will fail on a non-protected project.

google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
```diff
@@ -0,0 +1,391 @@
...
+
+  @Test
+  public void deniedCreateSubscription() {
+    try {
+      adminClient.createSubscription(subscription).get();
+    } catch (InterruptedException e) {
+      fail("Expected PERMISSION_DENIED StatusRuntimeException but got: " + e.toString());
+    } catch (ExecutionException e) {
+      Throwable thrown = e.getCause();
+      checkExceptionForVPCError((StatusRuntimeException) thrown);
```
dpcollins-google, line 272: Have you ever run this? Most of the exceptions thrown by the client libs are StatusException not StatusRuntimeException. Use ExtractStatus.toCanonical here and elsewhere to not have to care about this.

ORabbit: Yes in most cases I received `StatusRuntimeException`s except when calling Publish and Subscribe.
However opting to use `ExtractStatus.toCanonical` as it is much cleaner.

google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/it/ITVPCNegativeTest.java
ORabbit: See my comment above

PR #163, opened 2020-07-15
feat: Add the ability to use automatic subscriber assignment to the subscriber settings

dpcollins-google:

### DIFF
--- a/google-cloud-pubsublite/pom.xml
+++ b/google-cloud-pubsublite/pom.xml
@@ -196,6 +196,21 @@
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>flatten-maven-plugin</artifactId>
       </plugin>
+      <plugin>
+        <!--TODO: Remove before GA. -->
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>clirr-maven-plugin</artifactId>
+        <configuration>
+          <failOnError>false</failOnError>
+        </configuration>
+        <executions>
+          <execution>
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 </project>
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/SubscriberSettings.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/SubscriberSettings.java
@@ -24,16 +24,20 @@ import com.google.cloud.pubsublite.SequencedMessage;
 import com.google.cloud.pubsublite.SubscriptionPath;
 import com.google.cloud.pubsublite.SubscriptionPaths;
 import com.google.cloud.pubsublite.cloudpubsub.internal.AckSetTrackerImpl;
+import com.google.cloud.pubsublite.cloudpubsub.internal.AssigningSubscriber;
 import com.google.cloud.pubsublite.cloudpubsub.internal.MultiPartitionSubscriber;
+import com.google.cloud.pubsublite.cloudpubsub.internal.PartitionSubscriberFactory;
 import com.google.cloud.pubsublite.cloudpubsub.internal.SinglePartitionSubscriber;
 import com.google.cloud.pubsublite.internal.Preconditions;
+import com.google.cloud.pubsublite.internal.wire.AssignerBuilder;
+import com.google.cloud.pubsublite.internal.wire.AssignerFactory;
 import com.google.cloud.pubsublite.internal.wire.CommitterBuilder;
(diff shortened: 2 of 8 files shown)


anguillanneuf:
Please fix the tests first.

samples/snippets/pom.xml
```diff
@@ -46,7 +46,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-pubsublite</artifactId>
-      <version>0.1.7</version>
+      <version>0.1.9-SNAPSHOT</version>
```
anguillanneuf, line 49: Please don't update this to SNAPSHOT as this will get mirrored in https://cloud.google.com/pubsub/lite/docs/reference/libraries

google-cloud-pubsublite/pom.xml
```diff
@@ -196,6 +196,21 @@
...
+          <failOnError>false</failOnError>
+        </configuration>
+        <executions>
+          <execution>
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
```
anguillanneuf, line 213: Are we ready to remove this? Looks like tests are failing.

dpcollins-google: No, this change requires introducing it.

dpcollins-google:
> Please fix the tests first.

The tests are broken due to "**[08:14:28][ERROR] Timed out waiting for GitHub to return mergeable status for the pull (PR). GitHub is taking too long to compute whether this PR is mergeable. Please retry the build in a few minutes. You can do this by attaching a kokoro:force-run label on your PR.**" which sounds like a kokoro issue.

samples/snippets/pom.xml
dpcollins-google: Autosynth doesn't seem to be updating this number? At least it didnt in release 1.8 which is confusing.

anguillanneuf: It may have something to do with most snippets folders usually don't require version updates / they use libraries bom. @chingor13 is that right? See history: https://github.com/googleapis/java-pubsublite/commits/master/samples/snippets/pom.xml How do we overwrite the update rule?

dpcollins-google:
.readme-partials.yaml updated.

anguillanneuf:
Samples update should come later, after the library has a new release that picks up the changes. If you want, you can split this into two. Or I can update the samples later.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/SubscriberSettings.java
```diff
@@ -113,18 +125,36 @@ Subscriber instantiate() throws StatusException {
...
+            assignerBuilder.setReceiver(receiver);
+            return assignerBuilder.build();
+          };
+      return new AssigningSubscriber(partitionSubscriberFactory, assignerFactory);
+    }
+
+    List<Subscriber> perPartitionSubscribers = new ArrayList<>();
+    for (Partition partition : partitions().get()) {
+      wireSubscriberBuilder.setPartition(partition);
+      wireCommitterBuilder.setPartition(partition);
```
hannahrogers-google, line 156: I don't think we need these lines anymore:

```
wireSubscriberBuilder.setPartition(partition);
wireCommitterBuilder.setPartition(partition);
```

This should be covered by the partitionSubscriberFactory, right?

hannahrogers-google:
> I would like @hannahrogers-google or someone else comment on if more tests need to be added before you merge.

I am ok with this change - I don't believe we have explicit unit tests for any of the Builders in the lite client library.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/SubscriberSettings.java
dpcollins-google: Nack. This is still needed in the case when the user explicitly provides a list of partitions.

hannahrogers-google: Agreed we still need to set the partition, but we already do this inside partitionSubscriberFactory [L130-131].  When we pass the partition to the partitionSubscriberFactory below, we will again set the partition on the wireSubscriberBuilder and wireCommitterBuilder.  Or am I missing something?

anguillanneuf:
@dpcollins-google Does this Do Not Merge label still apply?

dpcollins-google:
> @dpcollins-google Does this Do Not Merge label still apply?

Yes it does. We don't want to merge this until the backend is tested internally and we've updated (/ are updating) the docs.

PR #169, opened 2020-07-20
fix: Change AdminClient close exception to be StatusException.

### LINKED ISSUE #148: Better exception for AdminClient
The exception that needs handling from the autocloseable `try(AdminClient adminClient = AdminClient.create(adminClientSettings)` is `java.lang.Exception`, which is very generic.

Pub/Sub uses `java.io.IOException` for `TopicAdminClient` and `SubscriptionAdminClient` ([samples](https://github.com/googleapis/java-pubsub/tree/master/samples/snippets/src/main/java/pubsub)).

See original [comment](https://github.com/googleapis/java-pubsublite/pull/147#discussion_r444510854) by @kurtisvg.

dpcollins-google:
Fixes #148

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/AdminClient.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/AdminClient.java
@@ -146,4 +146,11 @@ public interface AdminClient extends BackgroundResource {
    *     status {@link io.grpc.Status.Code#NOT_FOUND}
    */
   ApiFuture<Void> deleteSubscription(SubscriptionPath path);
+
+  /**
+   * Tear down this admin client.
+   *
+   * @throws StatusException on a failure to properly terminate.
+   */
+  void close() throws StatusException;
 }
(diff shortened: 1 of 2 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/AdminClient.java
```diff
@@ -146,4 +146,11 @@ static AdminClient create(AdminClientSettings settings) throws StatusException {
    *     status {@link io.grpc.Status.Code#NOT_FOUND}
    */
   ApiFuture<Void> deleteSubscription(SubscriptionPath path);
+
+  /**
+   * Tear down this admin client.
+   *
+   * @throws StatusException on a failure to properly terminate.
+   */
+  void close() throws StatusException;
```
elharo, line 155: This is a breaking change that requires a major version update.

dpcollins-google: There are no major versions in a pre-v1 library. This change will not be submitted as it is superceded by #295. This is also only a breaking change for someone who overrides (not uses) AdminClient, which is not an intended workflow. The previous method signature was `void close() throws Exception` which this is a subset of.

PR #171, opened 2020-07-23
fix: Update synth.py and run synthtool to get new pubsublite sources.

dpcollins-google:
This has the side effect of generating the GAPIC shim. We should consider using this although its unclear if it would give us any benefit.


dpcollins-google:
> In libraries that have gone GA, breaking changes will require a major version bump.
> Even though we allow breaking changes in pre-GA libraries, it's not a great user experience that we want to promote, it shouldn't happen too much too frequently.

In working on this client library, every update to an autogenerated piece of code, whether it is protobufs or autovalue, triggers this check. I don't have a problem with saying "we will not have breaking changes to the user-visible api surface for GA libraries", the problem is the clirr check as it exists does not appear to be sensitive enough to detect those, instead firing on any change to the 'abstract classes' that really only have one implementation created by protobuf and autovalue. If there was a way to make it not trigger on those, I would have no problem leaving it enabled, but it has such a high false positive rate that I don't believe it is worthwhile.

PR #189, opened 2020-08-05
samples: update samples for automatic subscriber assignment

anguillanneuf:
Create PR after #163 is merged and published on Maven Central.


dpcollins-google:
This all LGTM. Going to get the auto-assignment stuff on maven central today, then this should be submittable. If I could figure out how to approve I would, but it won't seem to let me.

PR #226, opened 2020-08-26
feat: Add the TopicBacklogReader

palmere-google:
This class allows users to provide a partition -> start cursor mapping,
and aggregates the message statistics over multiple partitions.

### DIFF
--- a/pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/TopicBacklogReader.java
+++ b/pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/TopicBacklogReader.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.pubsublite.beam;
+
+import com.google.api.core.ApiFuture;
+import com.google.cloud.pubsublite.Offset;
+import com.google.cloud.pubsublite.Partition;
+import com.google.cloud.pubsublite.proto.ComputeMessageStatsResponse;
+import java.util.Map;
+
+/**
+ * The TopicBacklogReader is intended for clients who would like to use the TopicStats API to
+ * aggregate the backlog, or the distance between the current cursor and HEAD across multiple
+ * partitions within a subscription.
+ */
+public interface TopicBacklogReader {
+  /**
+   * Compute and aggregate message statistics for message between the provided start offset and HEAD
+   * for each partition.
+   *
+   * @param subscriptionState A map from partition to the current offset of the subscriber in a
+   *     given partition.
+   * @return a future with either an error or a ComputeMessageStatsResponse with the aggregated
+   *     statistics for messages in the backlog on success.
+   */
+  ApiFuture<ComputeMessageStatsResponse> computeMessageStats(
+      Map<Partition, Offset> subscriptionState);
+}
--- a/pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/TopicBacklogReaderImpl.java
+++ b/pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/TopicBacklogReaderImpl.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/TopicBacklogReaderImpl.java
```diff
@@ -0,0 +1,102 @@
...
+              ComputeMessageStatsResponse.newBuilder()
+                  .setMessageBytes(messageBytes)
+                  .setMessageCount(messageCount);
+          if (minPublishTime.isPresent()) {
+            builder.setMinimumPublishTime(toProtoTimestamp(minPublishTime.get()));
+          }
+          if (minEventTime.isPresent()) {
+            builder.setMinimumEventTime(toProtoTimestamp(minEventTime.get()));
+          }
+          return ApiFutures.immediateFuture(builder.build());
```
dpcollins-google, line 98: this should just be .transform above if the future you return is immediately done.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/TopicBacklogReader.java
```diff
@@ -0,0 +1,41 @@
+/*
```
dpcollins-google, line 1: This is probably not useful outside of dataflow, would you mind putting it in the beam directory?

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/TopicBacklogReaderImpl.java
```diff
@@ -0,0 +1,102 @@
...
+        subscriptionState.entrySet().stream()
+            .map(
+                e ->
+                    client.computeMessageStats(
+                        topicPath, e.getKey(), e.getValue(), Offset.of(Integer.MAX_VALUE)))
+            .collect(Collectors.toList());
+    return ApiFutures.transformAsync(
+        ApiFutures.allAsList(perPartitionFutures),
+        responses -> {
+          Optional<Instant> minPublishTime = Optional.empty();
```
dpcollins-google, line 72: You don't have to keep shuffling back and forth between instant and timestamp. Instead, use the protobuf Timestamps.compare method.

google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/internal/TopicBacklogReaderImplTest.java
```diff
@@ -0,0 +1,189 @@
...
+
+    ImmutableMap.Builder<Partition, Offset> builder = ImmutableMap.builder();
+    ApiFuture<ComputeMessageStatsResponse> future =
+        reader.computeMessageStats(
+            builder
+                .put(Partition.of(1), Offset.of(10))
+                .put(Partition.of(2), Offset.of(20))
+                .build());
+
+    assertThat(future.get()).isEqualTo(aggregate);
```
dpcollins-google, line 111: You'll have to change this when you put it in beam: they won't add truth as a dependency. Use the generic mockito asserts.

google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/internal/TopicBacklogReaderImplTest.java
```diff
@@ -0,0 +1,189 @@
...
+            topicPath, Partition.of(1), Offset.of(10), Offset.of(Integer.MAX_VALUE)))
+        .thenReturn(ApiFutures.immediateFuture(partition1));
+    when(mockClient.computeMessageStats(
+            topicPath, Partition.of(2), Offset.of(20), Offset.of(Integer.MAX_VALUE)))
+        .thenReturn(ApiFutures.immediateFuture(partition2));
+
+    ImmutableMap.Builder<Partition, Offset> builder = ImmutableMap.builder();
+    ApiFuture<ComputeMessageStatsResponse> future =
+        reader.computeMessageStats(
+            builder
```
dpcollins-google, line 138: Just use ImmutableMap.of

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/TopicBacklogReaderImpl.java
palmere-google: Oh, yeah, I had some issues with transform before I realized you could use a lambda instead of implementing the ApiFunction class, but this great

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/TopicBacklogReader.java
palmere-google: Moved, hopefully correctly

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/TopicBacklogReaderImpl.java
palmere-google: Ah, cool, done

google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/internal/TopicBacklogReaderImplTest.java
palmere-google: Oh, didn't know that existed - much better

google-cloud-pubsublite/src/test/java/com/google/cloud/pubsublite/internal/TopicBacklogReaderImplTest.java
palmere-google: I switched to using the stuff from

import static org.junit.Assert.*;

Is that what you meant?

PR #231, opened 2020-09-03
fix: Embed builder in path classes and make components directly accessible

dpcollins-google:


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/PartitionLookupUtils.java
```diff
@@ -32,9 +32,7 @@ private PartitionLookupUtils() {}
   public static int numPartitions(TopicPath topic) throws StatusException {
     try (AdminClient client =
         AdminClient.create(
-            AdminClientSettings.newBuilder()
-                .setRegion(topic.location().location().region())
-                .build())) {;
+            AdminClientSettings.newBuilder().setRegion(topic.location().region()).build())) {;
```
manuelmenzella-google, line 35: Please remove the spurious trailing semi-colon.

manuelmenzella-google (changes requested):

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/PartitionLookupUtils.java
dpcollins-google: Done.

manuelmenzella-google (approved):

PR #235, opened 2020-09-07
docs: Update samples for v0.3.0

dpcollins-google:


manuelmenzella-google (approved):
Tianzi may be interested in these changes.

.readme-partials.yaml
```diff
@@ -21,10 +21,10 @@ custom_content: |
   Integer partitions = 1;
 
   TopicPath topicPath =
-      TopicPaths.newBuilder()
-          .setProjectNumber(ProjectNumber.of(projectNumber))
-          .setZone(CloudZone.of(CloudRegion.of(cloudRegion), zoneId))
-          .setTopicName(TopicName.of(topicId))
+      TopicPath.newBuilder()
+          .setProject(ProjectNumber.of(projectNumber))
```
lesv, line 25: don't we use projectName everywhere else in cloud?

lesv (approved):
LGTM - for java-sample-reviewers
I like the changes to the API, but wonder about using the project Number instead of the more common projectName.
You might wish to check the warnings from CodeCov - as the changes it calls out as not being covered seem like important parts of the sample.

PR #276, opened 2020-10-02
feat: Implement interfaces and utilities needed for Pub/Sub Lite Kafka shim

dpcollins-google:
Also run mvn fmt:format


pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/KafkaExceptionUtils.java
```diff
@@ -0,0 +1,92 @@
...
+   * @param t A throwable to transform.
+   * @return The transformed exception suitable for re-throwing.
+   */
+  static RuntimeException toKafka(Throwable t) {
+    try {
+      throw t;
+    } catch (KafkaException | UnsupportedOperationException | IllegalStateException e) {
+      return e;
+    } catch (InterruptedException e) {
+      return new InterruptException(e);
```
manuelmenzella-google, line 85: Nit; just "return e;"? What's the difference?

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/SingleSubscriptionConsumer.java
```diff
@@ -0,0 +1,53 @@
...
+  ApiFuture<Void> commit(Map<Partition, Offset> commitOffsets);
+
+  void doSeek(Partition partition, SeekRequest request) throws KafkaException;
+
+  Optional<Long> position(Partition partition);
+
+  void close(Duration duration);
+
+  /**
+   * Cause the outstanding or next call to poll to throw a WakeupException. The consumer is left in
```
manuelmenzella-google, line 49: What happens if a poll() is in progress? Kafka allows waking up an in-progress poll.

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/KafkaExceptionUtils.java
```diff
@@ -0,0 +1,92 @@
...
+  private KafkaExceptionUtils() {}
+
+  static KafkaException toKafkaException(StatusException source) {
+    switch (source.getStatus().getCode()) {
+      case OK:
+        throw source.getStatus().asRuntimeException();
+      case ABORTED:
+        return new BrokerNotAvailableException("Aborted.", source);
+      case ALREADY_EXISTS:
+        return new TopicExistsException("", source);
```
manuelmenzella-google, line 40: I'd be inclined to just make this a raw KafkaException, since there is no guarantee that the NOT_FOUND error is for a topic (and not a subscription).

manuelmenzella-google (approved):

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/KafkaExceptionUtils.java
dpcollins-google: InterruptException derives from KafkaException and is unchecked- InterruptedException is the java stdlib one thrown when a thread is interrupted.

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/SingleSubscriptionConsumer.java
dpcollins-google: "the outstanding or next call to poll"

this will work.

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/KafkaExceptionUtils.java
dpcollins-google: Done.

PR #281, opened 2020-10-05
feat: Implement SingleSubscriptionConsumerImpl

dpcollins-google:
This is an API that can be used to implement the kafka Consumer API.

### DIFF
--- a/pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/SingleSubscriptionConsumerImpl.java
+++ b/pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/SingleSubscriptionConsumerImpl.java
@@ -1,0 +1,288 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.pubsublite.kafka;
+
+import static com.google.cloud.pubsublite.kafka.KafkaExceptionUtils.toKafka;
+
+import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutureCallback;
+import com.google.api.core.ApiFutures;
+import com.google.cloud.pubsublite.Offset;
+import com.google.cloud.pubsublite.Partition;
+import com.google.cloud.pubsublite.SequencedMessage;
+import com.google.cloud.pubsublite.TopicPath;
+import com.google.cloud.pubsublite.internal.CloseableMonitor;
+import com.google.cloud.pubsublite.internal.ExtractStatus;
+import com.google.cloud.pubsublite.internal.PullSubscriber;
+import com.google.cloud.pubsublite.internal.wire.Committer;
+import com.google.cloud.pubsublite.proto.SeekRequest;
+import com.google.cloud.pubsublite.proto.SeekRequest.NamedTarget;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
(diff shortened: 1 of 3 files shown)


pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/SingleSubscriptionConsumerImpl.java
```diff
@@ -0,0 +1,288 @@
...
+    try (CloseableMonitor.Hold h = monitor.enter()) {
+      return partitions.keySet();
+    }
+  }
+
+  @Override
+  public ConsumerRecords<byte[], byte[]> poll(Duration duration) {
+    Map<Partition, Queue<SequencedMessage>> partitionQueues = new HashMap<>();
+    try {
+      if (wakeupTriggered) throw new WakeupException();
```
manuelmenzella-google, line 137: Can you read this without the mutex?

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/SingleSubscriptionConsumerImpl.java
```diff
@@ -0,0 +1,288 @@
...
+      while (!duration.isZero()) {
+        Duration sleepFor = Collections.min(ImmutableList.of(duration, Duration.ofMillis(100)));
```
manuelmenzella-google, line 139: I'm worried that this will result in very high e2e latency for consumers that set high poll durations. IIUC, Kafka returns from Poll as soon as any data is available, and does not wait until "max.poll.records" records are available. Ideally, poll() would return as soon as any one partition has any data unless "fetch.min.bytes" is set to >1.

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/SingleSubscriptionConsumerImpl.java
```diff
@@ -0,0 +1,288 @@
...
+          partitions.forEach(
+              ExtractStatus.rethrowAsRuntime(
+                  (partition, state) -> {
+                    partitionQueues
+                        .computeIfAbsent(partition, x -> new ArrayDeque<>())
+                        .addAll(state.subscriber.pull());
+                  }));
+        }
+      }
+      try (CloseableMonitor.Hold h = monitor.enter()) {
```
manuelmenzella-google, line 153: Do we need this last read?

dpcollins-google: Yes. Added an explanatory comment as to why.

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/SingleSubscriptionConsumerImpl.java
dpcollins-google: Yes, I was misreading the behavior. I've restructured this to return as soon as any messages are available, PTAL.

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/SingleSubscriptionConsumerImpl.java
dpcollins-google: No. Fixed.

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/SingleSubscriptionConsumerImpl.java
manuelmenzella-google: This works well for cases where messages are already in the buffer when the client calls poll(), but it will still have high latency when messages are delivered during the 100ms sleep. I can think of some cases where this will matter: imagine messages being published uniformly every 10ms. Also, we need to remember that clients are used to single-digit milliseconds [latency](https://stackoverflow.com/a/35071082) for Kafka. Would it be possible to have some kind of notification when any of the SinglePartitionSubscribers has messages ready to be delivered?

manuelmenzella-google (commented):
I have a single comment that I would like to discuss regarding latency for Kafka subscribers. Everything else looks great, thanks Dan.

manuelmenzella-google (approved):

dpcollins-google:
> I have a single comment that I would like to discuss regarding latency for Kafka subscribers. Everything else looks great, thanks Dan.

Ack. As discussed offline, this requires a bit of a gross "register watch" type api to implement. I will add a FR for this behavior and lower the polling duration to 10 ms in a subsequent PR.

PR #287, opened 2020-10-08
feat: Implement PubsubLiteConsumer.

dpcollins-google:
This implements Kafka's Consumer API.

### DIFF
--- a/pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/PubsubLiteConsumer.java
+++ b/pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/PubsubLiteConsumer.java
@@ -62,7 +62,8 @@ import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.UnsupportedVersionException;
 
 /**
- * A class that uses a MinimalKafkaConsumer to remove the duplicate methods from the kafka consumer.
+ * A class that uses a SingleSubscriptionConsumer to remove the duplicate methods from the kafka
+ * consumer.
  *
  * <p>This also filters methods that Pub/Sub Lite will not implement.
  */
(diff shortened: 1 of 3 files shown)


pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/PubsubLiteConsumer.java
```diff
@@ -0,0 +1,555 @@
...
+import org.apache.kafka.clients.consumer.OffsetAndTimestamp;
+import org.apache.kafka.clients.consumer.OffsetCommitCallback;
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.PartitionInfo;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.errors.UnsupportedVersionException;
+
+/**
+ * A class that uses a MinimalKafkaConsumer to remove the duplicate methods from the kafka consumer.
```
manuelmenzella-google, line 65: What is a MinimalKafkaConsumer? I cannot find references to it.

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/PubsubLiteConsumer.java
```diff
@@ -0,0 +1,555 @@
...
+  public ConsumerRecords<byte[], byte[]> poll(long l) {
+    return poll(Duration.ofMillis(l));
+  }
+
+  @Override
+  public ConsumerRecords<byte[], byte[]> poll(Duration timeout) {
+    return requireValidConsumer().poll(timeout);
+  }
+
+  Map<Partition, Offset> checkAndTransform(Map<TopicPartition, OffsetAndMetadata> map) {
```
manuelmenzella-google, line 250: I would personally make these method names more descriptive. "checkAndTransform" doesn't really convey what this method is doing without looking at the return/argument types, which are not always immediately visible at the call sites.

There's others like this: toKafka, toDuration, etc, but I think those are more obvious.

manuelmenzella-google (approved):
Minor nits. Thanks.

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/PubsubLiteConsumer.java
dpcollins-google: Done.

pubsublite-kafka-shim/src/main/java/com/google/cloud/pubsublite/kafka/PubsubLiteConsumer.java
dpcollins-google: Done.

PR #295, opened 2020-10-16
feat: All changes required to change library to use ApiException and gapic internals

### LINKED ISSUE #148: Better exception for AdminClient

### LINKED ISSUE #180: Consider converting internals to use GAPIC layer
The internals are currently written using GRPC. We should investigate whether we can remove some of our code by converting this to use GAPIC, but we should not pursue this unless we can identify which pieces of code we can remove beforehand.

dpcollins-google:
This converts the internals of the client library to use GAPIC generated clients, as well as the api surface to raise ApiException as opposed to grpc StatusException.

fixes #180

fixes #148 

This does not throw RESOURCE_EXHAUSTED at high partition counts with the following code when removing RESOURCE_EXHAUSTED from the list of retryable errors:

```java 
public class PublisherTest {
  public static void main(String[] args) throws Exception {
    Publisher publisher = Publisher.create(PublisherSettings.newBuilder()
        .setTopicPath(TopicPath.parse(<400 partition topic>)).build());
    publisher.startAsync().awaitRunning();
    publisher.publish(PubsubMessage.newBuilder().setData(ByteString.copyFromUtf8("abc")).build()).get();
    publisher.stopAsync().awaitTerminated();
  }
}
```


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/CheckedApiException.java
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.pubsublite.internal;
+
+import com.google.api.gax.rpc.ApiException;
+import com.google.api.gax.rpc.StatusCode;
+import javax.annotation.Nullable;
+
+public class CheckedApiException extends Exception {
```
manuelmenzella-google, line 23: Nit: can you add a comment on why we need this? IIUC, this just wraps an ApiException but makes it checked for visibility.

manuelmenzella-google (approved):

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/CheckedApiException.java
dpcollins-google: We don't 'need' this. Its to help with internal correctness, so that we have to handle errors that might arise. IMO, having ApiException inherit from RuntimeException is a mistake, but its what we have to work with.

dpcollins-google: Done.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/Subscriber.java
```diff
@@ -16,12 +16,12 @@
 
 package com.google.cloud.pubsublite.cloudpubsub;
 
+import com.google.api.gax.rpc.ApiException;
 import com.google.cloud.pubsub.v1.SubscriberInterface;
-import io.grpc.StatusException;
 
 public interface Subscriber extends SubscriberInterface {
```
anguillanneuf, line 22: Add description.

anguillanneuf (changes requested):
Thanks for making this PR. I didn't review any of the tests. Here are some issues I found:

1. This comment about setting stub on the publisher needs an update: 
   https://github.com/googleapis/java-pubsublite/blob/7b6bd0545e49d7c7066764776b2cd2501b73a22f/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PublisherBuilder.java#L56
1. I'm not sure if you intended `google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/testing/UnitTestExamples.java` to be there.
1. Sporadic class descriptions in `google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/`.
1. Changes in `pubsublite-beam-io/` and `pubsublite-kafka-shim/` should be their own PRs. 
1. Changes to `grpc-google-cloud-pubsublite-v1/` should come from synthtool pulling from the source. 
1. Update samples in a later PR.
1. Also please resolve the conflicts with main.

I haven't tried the code to see how they impact samples myself. I can do that next week.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/ApiBackgroundResource.java
```diff
@@ -17,14 +17,14 @@
...
 import com.google.api.gax.core.BackgroundResource;
-import io.grpc.StatusException;
+import com.google.api.gax.rpc.ApiException;
 
 public interface ApiBackgroundResource extends BackgroundResource {
   /**
    * Tear down this resource.
    *
-   * @throws StatusException on a failure to properly terminate.
+   * @throws ApiException on a failure to properly terminate.
```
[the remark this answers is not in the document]
dpcollins-google: Nack. `@throws` does this automatically in javadoc

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/AssignerImpl.java
```diff
@@ -41,29 +44,42 @@
...
+
   @Override
   public void triggerReinitialize() {
     try (CloseableMonitor.Hold h = monitor.enter()) {
       connection.reinitialize();
     }
   }
 
-  private static Set<Partition> toSet(PartitionAssignment assignment) throws StatusException {
+  private static Set<Partition> toSet(PartitionAssignment assignment) {
```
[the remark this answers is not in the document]
dpcollins-google: ApiException is unchecked. I introduced CheckedApiException for internal usage because I believe this to be a mistake, but it does not need to be declared.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/Subscriber.java
dpcollins-google: Done.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/TrivialProxyService.java
```diff
@@ -17,17 +17,19 @@
...
-import io.grpc.StatusException;
+import com.google.api.gax.rpc.ApiException;
 import java.util.Arrays;
 import java.util.Collection;
 
+/** A ProxyService that just wraps all ApiService methods. */
 public class TrivialProxyService extends ProxyService {
-  public TrivialProxyService(ApiService... services) throws StatusException {
-    this(Arrays.asList(services));
+  protected <T extends ApiService> TrivialProxyService(Collection<T> services) throws ApiException {
```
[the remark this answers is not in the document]
dpcollins-google: Done.

dpcollins-google:
> Thanks for making this PR. I didn't review any of the tests. Here are some issues I found:

Replies to each point below.

1) Done.
2) Yes, since this needs to be accessed from other modules.
3) This is not a public API, thus it can be put off until later as the only users of these classes would be maintainers.
4) No, since they would not compile otherwise.
5) I've reverted these. They were unintentional.
6) No, we've discussed this previously. For the integration tests to pass as currently set up, samples changes need to be made in the same PR, since the blocking presubmit is the snapshot one.
7) Done.

anguillanneuf (approved):
3. It's in the public [documentation](https://googleapis.dev/java/google-cloud-pubsublite/latest/index.html), either we add more documentation there, or we choose to hide it. Can be in a later PR. 
6. Ok. Make sure to make a new release soon after this PR is merged then.

PR #323, opened 2020-10-22
docs: Provide worked Kafka example with beam

dpcollins-google:
This is intended to be a worked example: It may or may not be submitted or could be moved depending on tianzi's thoughts.

### DIFF
--- a/samples/snippets/src/main/java/pubsublite/kafka/KafkaIOReadExample.java
+++ b/samples/snippets/src/main/java/pubsublite/kafka/KafkaIOReadExample.java
@@ -1,0 +1,118 @@
+package com.google.cloud.pubsublite.demo;
+
+import com.google.cloud.pubsublite.CloudZone;
+import com.google.cloud.pubsublite.ProjectNumber;
+import com.google.cloud.pubsublite.SubscriptionName;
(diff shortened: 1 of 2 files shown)


samples/snippets/src/main/java/pubsublite/kafka/KafkaIOReadExample.java
```diff
@@ -0,0 +1,118 @@
+package com.google.cloud.pubsublite.demo;
```
lesv, line 1: Shouldn't use `com.google.*` for sample / demo package names.  `com.example.*` or `productname.xxx`

lesv: We don't want customers accidently publishing w/ com.google

lesv (changes requested):
I'll come back to this after my next meeting, I don't see a pom.xml, so I wonder a bit.  Also, it looks like there might be a line length issue.

lesv:
Ah - I see checkstyle has failed, please fix those issues and then ping me.

dpcollins-google:
Closing this, it was never intended to be merged.

PR #370, opened 2020-11-16
feat: Implement per-server memory limiter for PubsubLiteIO

dpcollins-google:
This limiter is predicated on the assumptions that:
1) UnboundedSources have split() called at least once in their lifetime
2) UnboundedSources.split may return any amount of splits, even more than requested.
3) The per-partition memory requirements are high (multiple seconds of buffering) but splits will eventually be assigned to new machines if they are artificially restricted due to increasing backlog

It refreshes each individual reader on a machine every 5 minutes, as the memory limiter approaches an even distribution after multiple rounds of all leases being created and destroyed, and splits the IO into one source per partition to ensure that these refreshes are independent.

### DIFF
--- a/pubsublite-beam-io/README.md
+++ b/pubsublite-beam-io/README.md
@@ -65,5 +65,5 @@
     ```
 
 1. If operating in a memory-constrained environment, making your pipeline
-options inherit from `PubsubLitePipelineOptions` will put a 1GiB per
-machine limit on the accounted memory size of messages.
+options inherit from `PubsubLitePipelineOptions` will put a customizable (1GiB
+per machine by default) limit on the accounted memory size of messages.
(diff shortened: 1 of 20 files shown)


pubsublite-beam-io/README.md
```diff
@@ -58,8 +58,12 @@
...
             .setFlowControlSettings(FlowControlSettings.builder()
-                .setBytesOutstanding(100_000_000)  // 100 MB
+                .setBytesOutstanding(50_000_000)  // 50 MB
                 .setMessagesOutstanding(Long.MAX_VALUE)
                 .build())
             .build()));
     ```
+
+1. If operating in a memory-constrained environment, making your pipeline
+options inherit from `PubsubLitePipelineOptions` will put a 1GiB per
```
dpmills, line 68: This isn't the normal way to use options.  Just make the limit unset by default and let users set the limit if they want.  If you want to have a default value that's easy to use, have two options, one bool that turns on the limiter, and one long that sets the limit

dpcollins-google: Done.

dpcollins-google:
Closing, going to implement this in beam instead.

PR #386, opened 2020-11-20
feat: adds functions to BufferingPullSubscriber

jiangmichaellll:
This adds two methods to BufferingPullSubscriber so it would be easier to implement Spark's [ContinuousInputPartitionReader](http://spark.apache.org/docs/2.4.1/api/java/org/apache/spark/sql/sources/v2/reader/streaming/ContinuousInputPartitionReader.html)'s get and next.

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/BufferingPullSubscriber.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/BufferingPullSubscriber.java
@@ -113,6 +113,29 @@ public class BufferingPullSubscriber implements PullSubscriber<SequencedMessage>
   }
 
   @Override
+  public synchronized Optional<SequencedMessage> pullOne() throws CheckedApiException {
+    if (error.isPresent()) {
+      throw error.get();
+    }
+    if (messages.isEmpty()) {
+      return Optional.empty();
+    }
+    SequencedMessage message = messages.pop();
+    underlying.allowFlow(
+        FlowControlRequest.newBuilder()
+            .setAllowedBytes(message.byteSize())
+            .setAllowedMessages(1)
+            .build());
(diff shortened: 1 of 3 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/BufferingPullSubscriber.java
```diff
@@ -112,6 +112,29 @@ private synchronized void addMessages(Collection<SequencedMessage> new_messages)
     return ImmutableList.copyOf(collection);
   }
 
+  @Override
```
dpcollins-google, line 115: Can't you implement these on the user side, by just calling pull() then testing for emptiness (hasNext, although this is not exactly what spark wants for hasNext I don't think).

dpcollins-google:
These do not seem to implement the correct spark behavior, "Proceed to next record, returns false if there is no more records.". Just because there are none currently buffered does not mean that none are available in the range spark requested.

jiangmichaellll:
It's not strictly equal concepts (ie buffer isn't empty vs none are available), however, for continuous processing, what matters is if it has message to consume right at the moment (buffer isn't empty) instead of if it exists in PSL. Even when messages are available in PSL but somehow due to issues subscriber could not get it, from spark pov, it's no next message.

https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/continuous/ContinuousQueuedDataReader.scala#L142-L155 this is the place next(), get() and getOffset() are called. The process is periodically run, so it ensures when the message arrives in local buffer it will get picked up. 

If we implement by if message are available in PSL, the poll logic will need to be blocking to wait for message actually delivered in local buffer, which I think is worse.

jiangmichaellll:
Going to close this in favor of creating another layer of message queue.

PR #390, opened 2020-11-28
Sdf conversion staging

dpcollins-google:


dpcollins-google:
Merging this into beam mainline instead.

PR #426, opened 2020-12-14
feat: Spark micro batch processing

jiangmichaellll:
Implementation of spark micro batch processing. Many parts are similar to continuous processing.


pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslDataSource.java
```diff
@@ -78,4 +83,62 @@ public ContinuousReader createContinuousReader(
...
+    PslDataSourceOptions pslDataSourceOptions =
+        PslDataSourceOptions.fromSparkDataSourceOptions(options);
+    CursorClient cursorClient = pslDataSourceOptions.newCursorClient();
+    AdminClient adminClient = pslDataSourceOptions.newAdminClient();
+    SubscriptionPath subscriptionPath = pslDataSourceOptions.subscriptionPath();
+    long topicPartitionCount;
+    TopicPath topicPath;
+    try {
+      topicPath = TopicPath.parse(adminClient.getSubscription(subscriptionPath).get().getTopic());
+      topicPartitionCount = adminClient.getTopicPartitionCount(topicPath).get();
```
dpcollins-google, line 104: Please use the partition count utilities for this purpose.

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchInputPartition.java
```diff
@@ -0,0 +1,77 @@
...
+  @Override
+  public InputPartitionReader createPartitionReader() {
+    PslPartitionOffset pslPartitionOffset = PslSparkUtils.toPslPartitionOffset(startOffset);
+
+    BlockingPullSubscriber subscriber;
+    try {
+      subscriber =
+          new BlockingPullSubscriberImpl(
+              // TODO(jiangmichael): Pass credentials settings here.
+              (consumer) ->
```
dpcollins-google, line 59: nit. I'd push this up the stack a bit. Pass in a SerializableSupplier<BlockingPullSubscriber> instead, then you don't need to think about credentials or the start offset here at all.

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchInputPartitionReader.java
```diff
@@ -0,0 +1,98 @@
...
+        subscriber.onData().get(SUBSCRIBER_PULL_TIMEOUT.getSeconds(), TimeUnit.SECONDS);
+        msg = subscriber.messageIfAvailable();
+        break;
+      } catch (TimeoutException e) {
+        // NOTE this would only happen when a broker abnormally crashes and fails to
+        // write the unused reserved offsets back to storage. In this case, the
+        // spark pipeline could get stuck forever until a new message is published at
+        // head.
+        log.atWarning().log(
+            "Unable to get any messages in last " + SUBSCRIBER_PULL_TIMEOUT.toString());
```
dpcollins-google, line 69: You don't need to call "toString"

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchInputPartitionReader.java
```diff
@@ -0,0 +1,98 @@
...
+
+public class PslMicroBatchInputPartitionReader implements InputPartitionReader<InternalRow> {
+  private static final GoogleLogger log = GoogleLogger.forEnclosingClass();
+
+  private static final Duration SUBSCRIBER_PULL_TIMEOUT = Duration.ofSeconds(10);
+
+  private final SubscriptionPath subscriptionPath;
+  private final SparkPartitionOffset endOffset;
+  private final BlockingPullSubscriber subscriber;
+  private SequencedMessage currentMsg = null;
```
dpcollins-google, line 39: sanity check: None of this needs to be thread safe?

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchInputPartitionReader.java
```diff
@@ -0,0 +1,98 @@
...
+      return false;
+    }
+    Optional<SequencedMessage> msg;
+    while (true) {
+      try {
+        subscriber.onData().get(SUBSCRIBER_PULL_TIMEOUT.getSeconds(), TimeUnit.SECONDS);
+        msg = subscriber.messageIfAvailable();
+        break;
+      } catch (TimeoutException e) {
+        // NOTE this would only happen when a broker abnormally crashes and fails to
```
dpcollins-google, line 64: Remove this note. It doesn't belong in a public github, and its not the only time this would happen

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchInputPartitionReader.java
```diff
@@ -0,0 +1,98 @@
...
+        // write the unused reserved offsets back to storage. In this case, the
+        // spark pipeline could get stuck forever until a new message is published at
+        // head.
+        log.atWarning().log(
+            "Unable to get any messages in last " + SUBSCRIBER_PULL_TIMEOUT.toString());
+      } catch (Throwable t) {
+        throw new IllegalStateException("Failed to retrieve messages.", t);
+      }
+    }
+    // since next() is only called one at a time, we are sure that the message is
```
dpcollins-google, line 74: nit 'is only called on one thread at a time'

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchInputPartitionReader.java
```diff
@@ -0,0 +1,98 @@
...
+    // available to this thread.
+    assert msg.isPresent();
+    currentMsg = msg.get();
+    if (currentMsg.offset().value() >= endOffset.offset()) {
+      batchFulfilled = true;
```
dpcollins-google, line 79: should you return 'false' in this case and not set currentMsg?

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchInputPartitionReader.java
dpcollins-google, line 39: @Nullable

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchReader.java
```diff
@@ -0,0 +1,154 @@
...
+      assert SparkSourceOffset.class.isAssignableFrom(start.get().getClass())
+          : "start offset is not assignable to PslSourceOffset.";
+      startOffset = (SparkSourceOffset) start.get();
+    } else {
+      try {
+        Map<Partition, com.google.cloud.pubsublite.Offset> pslSourceOffsetMap = new HashMap<>();
+        for (int i = 0; i < topicPartitionCount; i++) {
+          pslSourceOffsetMap.put(Partition.of(i), com.google.cloud.pubsublite.Offset.of(0));
+        }
+        cursorClient
```
dpcollins-google, line 79: this method will not return a value if the offset for a partition has never been set.

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchReader.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  @Override
+  public void setOffsetRange(Optional<Offset> start, Optional<Offset> end) {
+    if (start.isPresent()) {
+      assert SparkSourceOffset.class.isAssignableFrom(start.get().getClass())
+          : "start offset is not assignable to PslSourceOffset.";
+      startOffset = (SparkSourceOffset) start.get();
+    } else {
+      try {
```
dpcollins-google, line 74: nit. make a method

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchReader.java
```diff
@@ -0,0 +1,154 @@
...
+public class PslMicroBatchReader implements MicroBatchReader {
+
+  private final CursorClient cursorClient;
+  private final HeadOffsetReader headOffsetReader;
+  private final MultiPartitionCommitter committer;
+  private final SubscriptionPath subscriptionPath;
+  private final TopicPath topicPath;
+  private final FlowControlSettings flowControlSettings;
+  private final long topicPartitionCount;
+  private SparkSourceOffset startOffset;
```
dpcollins-google, line 47: if these are nullable, annotate them as such and initialize them explicitly to null

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchReader.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  @Override
+  public List<InputPartition<InternalRow>> planInputPartitions() {
+    return startOffset.getPartitionOffsetMap().values().stream()
+        .map(
+            v -> {
+              SparkPartitionOffset endPartitionOffset =
+                  endOffset.getPartitionOffsetMap().get(v.partition());
+              if (Objects.equals(v, endPartitionOffset)) {
```
dpcollins-google, line 144: v.equals

dpcollins-google (changes requested):

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchInputPartitionReader.java
jiangmichaellll: No it's similar as continuous mode, that they are called serially by one thread.

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchInputPartitionReader.java
jiangmichaellll: the endoffset is a SparkPartitionOffset, and for spark, it's closed endoffset. If we return false here, the call site won't call get() so we will miss the last message in the batch. I guess what you are saying is that when > it should be return false while when = it should return true. I will modify to that.

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslMicroBatchReader.java
jiangmichaellll: Yea, I noticed that, that's why I have just above to get partition count and supply default value.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberFactory.java
```diff
@@ -19,9 +19,11 @@
 import com.google.api.gax.rpc.ApiException;
 import com.google.cloud.pubsublite.SequencedMessage;
 import com.google.common.collect.ImmutableList;
+import java.io.Serializable;
 import java.util.function.Consumer;
 
-public interface SubscriberFactory {
+@FunctionalInterface
```
dpcollins-google, line 25: not needed

pubsublite-spark-sql-streaming/src/main/java/com/google/cloud/pubsublite/spark/PslDataSource.java
```diff
@@ -131,14 +125,13 @@ public PslSourceOffset getHeadOffset(TopicPath topic) {
...
-    return new PslMicroBatchReader(
-        cursorClient,
-        headOffsetReader,
-        committer,
-        subscriptionPath,
-        topicPath,
-        Objects.requireNonNull(pslDataSourceOptions.flowControlSettings()),
-        topicPartitionCount);
+    try {
+      TopicPath tp =
```
dpcollins-google, line 129: You should probably only look this up once so you don't hit rate limits on this request, and pass the topic path into the caller of this?

dpcollins-google (approved):

PR #432, opened 2020-12-17
feat: Multiple fixes to make continuous mode work

jiangmichaellll:
This brings the continuous mode working with non-uber jar solution. Uber jar still won't work.

Mainly:
1. Lower scala version to 2.11 since it's what's provided in spark 2.4.7 in classpath
2. Fixes the conversion from java types into spark types
3. Added maven shade plugin for uber jar generation


pubsublite-spark-sql-streaming/pom.xml
```diff
@@ -203,6 +168,49 @@
           </execution>
         </executions>
       </plugin>
```
dpcollins-google, line 170: nit. you can probably get rid of this, just click on the "m" in intellij and run `mvn fmt:format`

dpcollins-google (approved):

PR #445, opened 2021-01-08
feat: Update spark connector pom.xml

jiangmichaellll:
1. use auto service instead of auto service annotation in the processor. https://screenshot.googleplex.com/AZCi8QjvdQ9AdC2
2. Add relocations to resolve grpc-netty-shaded issues.

### DIFF
--- a/pubsublite-spark-sql-streaming/pom.xml
+++ b/pubsublite-spark-sql-streaming/pom.xml
@@ -21,8 +21,6 @@
     <scala.version>2.11.12</scala.version>
     <scala.version.short>2.11</scala.version.short>
     <spark.version>2.4.7</spark.version>
-    <grpc.netty.version>1.30.2</grpc.netty.version>
-    <netty.tcnative.version>2.0.34.Final</netty.tcnative.version>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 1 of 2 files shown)


jiangmichaellll:
I am going to try another suggestion from dataproc team. This PR makes it working in dataproc but restrains the fat jar with specific versions of tcnative that might not work with other images.

pubsublite-spark-sql-streaming/pom.xml
```diff
@@ -19,24 +19,28 @@
     <maven.compiler.target>1.8</maven.compiler.target>
     <encoding>UTF-8</encoding>
     <scala.version>2.11.12</scala.version>
+    <scala.version.short>2.11</scala.version.short>
+    <spark.version>2.4.7</spark.version>
+    <grpc.netty.version>1.30.2</grpc.netty.version>
+    <netty.tcnative.version>2.0.34.Final</netty.tcnative.version>
```
medb, line 25: Are these versions used anywhere? If you are using `grpc-netty-shaded` than probably you do not need to set these explicitly.

pubsublite-spark-sql-streaming/pom.xml
```diff
@@ -176,6 +181,36 @@
                     <include>com.google.common.**</include>
                   </includes>
                 </relocation>
+                <!-- Take special care of grpc-netty-shaded, it uses the package
+                     io.grpc.netty.shaded.io.grpc.netty, which will cause the
+                     ServicesResourceTransformer to replace both occurrences of io.grpc -->
+                <relocation>
+                  <pattern>io.grpc.netty.shaded</pattern>
+                  <shadedPattern>
+                    com.google.cloud.hadoop.repackaged.gcs.io.grpc.netty.shaded
```
medb, line 190: You need to change `com.google.cloud.hadoop.repackaged.gcs` here and below to `com.google.cloud.pubsublite.repackaged` or something like this so it will be unique to this project.

medb (changes requested):

pubsublite-spark-sql-streaming/pom.xml
jiangmichaellll: Removed.

medb (approved):

dpcollins-google (approved):

PR #449, opened 2021-01-12
feat: fix bug where blockingpullsubscriber doesn't refill flowcontrol quota.

jiangmichaellll:
As titled.


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/BlockingPullSubscriberImpl.java
```diff
@@ -113,7 +112,13 @@ private synchronized void addMessages(Collection<SequencedMessage> new_messages)
     if (messages.isEmpty()) {
       return Optional.empty();
     }
-    return Optional.of(Objects.requireNonNull(messages.pollFirst()));
+    SequencedMessage msg = messages.pollFirst();
```
dpcollins-google, line 115: nit. removefirst

dpcollins-google (approved):

PR #467, opened 2021-01-26
feat: Remove spark connector as a module

jiangmichaellll:
Since it's moved into java-pubsublite-spark and no longer maintained here.


anguillanneuf (approved):
Thanks for this. The source code now lives in https://github.com/googleapis/java-pubsublite-spark

PR #479, opened 2021-02-01
feat: remove beam I/O from pubsublite github and redirect to beam version instead

dpcollins-google:


anguillanneuf:
Let's make sure the Beam version works before removing. 
We shall use `feat` to prefix the commit message this time so it shows up in the release notes.

anguillanneuf (approved):
Tested a simple Dataflow job using the beam library, okay to redirect users over there.

PR #482, opened 2021-02-02
feat: Change message id on PubsubMessages to be an encoded MessageMetadata and rename PublishMetadata to MessageMetadata

### LINKED ISSUE #471: Support per partition MessageReceiver
**Is your feature request related to a problem? Please describe.**
There's currently no easy way to know from which partition a message come from while using the `Subscriber` interface. This prevent to implement more complex control flow on a per partition basis. e.g.: We might want to stop processing new messages if there's a single outstanding message with an offset smaller than current offset minus X. In addition, there's no way to determine a unique id as only the offset is available. Ideally the publisher should include an id as an attribute, but there's use case where distinguishing between a message that have been publish more than once vs a message that have been deliver more than once can be useful.

**Describe the solution you'd like**
Modify `SubcriberSettings.Builder` to add the method `setReceiverFactory(ApiFunction<Partition, MessageReceiver> receiverFactory)`. The method `setReceiver(MessageReceiver receiver)` should simply call this method, ignoring the `Partition` parameter.

**Describe alternatives you've considered**

Using `AssigningSubscriber` by giving `PartitionSubscriberFactory` that simply build a `Subscriber` from a `SubcriberSettings` where `partitions` have been set to a list of one element. Those are in the `com.google.cloud.pubsublite.cloudpubsub.internal` so it doesn't seem to be something that should be used outside the library.

dpcollins-google:
PublishMetadata is both the partition and offset, and is an ID suitable for deduplication.

Fixes #471


samples/snippets/src/main/java/pubsublite/SubscriberExample.java
```diff
@@ -68,7 +69,7 @@ public static void subscriberExample(
 
     MessageReceiver receiver =
         (PubsubMessage message, AckReplyConsumer consumer) -> {
-          System.out.println("Id : " + message.getMessageId());
+          System.out.println("Id : " + PublishMetadata.decode(message.getMessageId()));
```
manuelmenzella-google, line 72: I think this is a good idea, and I'm glad you are exposing the partition through me message ID. I would prefer if we didn't parse this into a *Publish* metadata, since the message ID is not really only the concern of the publisher. What if we called this MessageId?

Then we'd have:

`MessageId.decode(message.getMessageId());`

manuelmenzella-google (changes requested):

samples/snippets/src/main/java/pubsublite/SubscriberExample.java
dpcollins-google: Discussed offline, renaming this type will probably have to be done separately since its public. We'll discuss renaming out of band, and can do so at a later date.

manuelmenzella-google (approved):

anguillanneuf (approved):
LGTM. 

You don't really need to change `README.md` as any change in `.readme-partials.yaml` will trigger an update in `README.md`.

PR #496, opened 2021-02-05
fix: Rebatch messages when restarting a publish stream

dpcollins-google:
This prevents publishers which are network isolated for a while from appearing disconnected when the server attempts to acknowledge messages.

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/Constants.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/Constants.java
@@ -31,7 +31,7 @@ public class Constants {
           .build();
 
   public static final long MAX_PUBLISH_BATCH_COUNT = 1_000;
-  public static final long MAX_PUBLISH_BATCH_BYTES = 3_500_000;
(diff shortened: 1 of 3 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/Constants.java
```diff
@@ -31,7 +31,7 @@
           .build();
 
   public static final long MAX_PUBLISH_BATCH_COUNT = 1_000;
-  public static final long MAX_PUBLISH_BATCH_BYTES = 3_500_000;
+  public static final long MAX_PUBLISH_BATCH_BYTES = 1024 * 1024 * 7 / 4;  // 3.5 MiB
```
manuelmenzella-google, line 34: I think you want 7/2, not 7/4. This is 1.75 MiB.

dpcollins-google: Good catch, fixed.

hannahrogers-google (approved):

PR #513, opened 2021-02-23
feat: Re-fork the beam repo from upstream

dpcollins-google:
The beam version has changes (notably the use of bundle finalizers) which make it impossible to use with dataflow at present.


manuelmenzella-google (approved):

anguillanneuf (changes requested):
@dpcollins-google Why do we need to move the code back here? Is it possible to continue developing it upstream?

PR #545, opened 2021-03-09
feat: Adding ability to create a subscription at HEAD

hannahrogers-google:


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/AdminClient.java
```diff
@@ -108,6 +119,17 @@ static AdminClient create(AdminClientSettings settings) throws ApiException {
    */
   ApiFuture<Subscription> createSubscription(Subscription subscription);
```
dpcollins-google, line 120: `default ApiFuture<Subscription> createSubscription(Subscription subscription) { createSubscription(subscription, OffsetLocation.END); }`

dpcollins-google: Also modify the comment to make it clear where this creates the subscription

dpcollins-google (changes requested):
LGTM after fixes. Ping me once fixed

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/AdminClient.java
hannahrogers-google: Done. This is ready for another look. Thanks!

dpcollins-google (approved):

PR #563, opened 2021-03-29
fix: Set batching settings on SinglePartitionPublisher

dpcollins-google:
Writing to a Lite topic will work with this fix.
But a pipeline that writes a bounded source to a Lite topic doesn't exit.

### DIFF
--- a/pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/Publishers.java
+++ b/pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/Publishers.java
@@ -26,6 +26,7 @@ import com.google.cloud.pubsublite.AdminClient;
 import com.google.cloud.pubsublite.AdminClientSettings;
 import com.google.cloud.pubsublite.MessageMetadata;
 import com.google.cloud.pubsublite.Partition;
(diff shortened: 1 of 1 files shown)


anguillanneuf (approved):
Tested with a simple example like below and it worked, but the pipeline doesn't exit. @dpcollins-google is aware and look into it a bit later.

```java
import com.google.cloud.pubsublite.Message;
import com.google.cloud.pubsublite.proto.PubSubMessage;
...
    pipeline
        .apply(Create.of("Hello!", "Hello again"))
        .apply(ParDo.of(new DoFn<String, PubSubMessage>() {
          @ProcessElement
          public void processElement(ProcessContext context) {
            ImmutableListMultimap<String, ByteString> multimap =
                ImmutableListMultimap.of("Year", ByteString.copyFromUtf8("2021"));
            Message message = Message.builder()
                .setData(ByteString.copyFromUtf8(context.element()))
                .setAttributes(multimap)
                //.setEventTime(fromMillis(System.currentTimeMillis()))
                .build();
            LOG.info("Created: " + Instant.now() + message);
            context.output(message.toProto());
          }
        }))
        .apply("Write to Pub/Sub Lite", PubsubLiteIO.write(publisherOptions));
```

PR #565, opened 2021-03-30
fix: Centralize alarm executor creation for Pub/Sub Lite

dpcollins-google:
Also prevent creation of per-object thread pools which are largely unused and may prevent shutdown if not using daemon threads.


anguillanneuf (approved):
I tested the fix with a simple example to write two messages to a Lite topic. Shutdown worked.

PR #618, opened 2021-04-27
feat: Compute cursors for publish and event timestamps

tmdiep:
Adds methods to the internal TopicStatsClient for computing cursors for publish or event timestamps.


palmere-google (approved):
This looks great!

PR #653, opened 2021-05-26
refactor: retrying connection updates

tmdiep:
Preparation for admin seek. This PR is a no-op.

Refactors `RetryingConnectionImpl`:

* Accepts a `InitialRequestProvider` instead of a static initial request proto. This will allow the initial subscribe request to change for stream reconnects (seeks to cursor).
* Accepts a `ResetHandler` and notifies it when the stream `RESET` signal is detected.


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/RetryingConnectionImpl.java
```diff
@@ -171,6 +178,14 @@ public final void onError(Throwable t) {
       setPermanentError(statusOr.get());
       return;
     }
+    if (ResetSignal.isResetSignal(statusOr.get())) {
```
dpcollins-google, line 181: Can you instead modify the triggerReinitialize method to pass the previous failure status? Then you can isolate this logic to the subscriber which needs to handle it.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/RetryingConnectionImpl.java
```diff
@@ -95,7 +101,8 @@ protected void doStart() {
   public void reinitialize() {
     try (CloseableMonitor.Hold h = connectionMonitor.enter()) {
       if (completed) return;
-      currentConnection = connectionFactory.New(streamFactory, this, initialRequest);
+      lastInitialRequest = initialRequestProvider.getInitialRequest();
```
dpcollins-google, line 104: proposal: Instead, make this method `public void reinitialize(StreamRequestT initialRequest)`. Then you prevent adding another upcall.

dpcollins-google (changes requested):

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/RetryingConnectionImpl.java
tmdiep: Done

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/RetryingConnectionImpl.java
tmdiep: Done.

However, I had to move the first `reinitialize` call from `RetryingConnectionImpl.doStart()` up to the stream impl `start()` methods because they store the initial request. I think this is more intuitive anyway, even though there is some duplication.

Also reordered some fields and methods in the stream impl classes, for consistency.

tmdiep: > However, I had to move the first reinitialize call from RetryingConnectionImpl.doStart() up to the stream impl start() methods because they store the initial request.

Actually, just realized that this will cause all stream initialization during startup to be blocking. Moved the first reinitialize() call back into RetryingConnectionImpl and passed the first initial request to the constructor.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/ResetSignal.java
```diff
@@ -0,0 +1,55 @@
...
+import io.grpc.protobuf.StatusProto;
+
+// Pub/Sub Lite's stream RESET signal is sent by the server to instruct the client to reset the
+// stream state.
+public final class ResetSignal {
+  private static final String REASON = "RESET";
+  private static final String DOMAIN = "pubsublite.googleapis.com";
+
+  public static boolean isResetSignal(CheckedApiException checkedApiException) {
+    if (!ErrorCodes.IsRetryableForStreams(checkedApiException.code())) {
```
dpcollins-google, line 33: can you be more specific that the reset signal is an ABORTED code?

dpcollins-google: Actually this is fine as-is.

dpcollins-google (approved):
Its a bit strange that the ResetSignal code is here, I'd weakly prefer it associated with non-refactor changes.

tmdiep:
> Its a bit strange that the ResetSignal code is here, I'd weakly prefer it associated with non-refactor changes.

Yeah, it was integrated in the first revision of this PR. But after refactoring, it became disjoint. Since it's already reviewed, I'll just include it.

Thanks Dan!

PR #662, opened 2021-06-03
feat: Ability to reset subscriber upon out of band seek

tmdiep:
Prepares for handling the `RESET` signal from the server, by:

* Discarding outstanding acks for delivered messages.
* Waiting for the committer to flush pending commits and receive the acknowledgment from the server.
* Then resetting subscriber state, including canceling any in-flight client seeks.

Notes:

* This PR is effectively a no-op as `SinglePartitionSubscriber.onSubscriberReset` returns false for now.
* The next PR will fully handle admin seeks by setting InitialSubscribeRequest.initial_location and implement `SinglePartitionSubscriber.onSubscriberReset`. Omitted to keep the size of the PR down.
* `CommitState.complete` had to be refactored to just throw (and not abort) because `CommitterImpl.waitUntilEmpty` should throw when there are excess commit responses, since the committer will shortly shut down.

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTracker.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTracker.java
@@ -25,7 +25,7 @@ interface AckSetTracker extends ApiService {
   // to add to the ack set. Must be called with strictly increasing offset messages.
   Runnable track(SequencedMessage message) throws CheckedApiException;
 
-  // Wait until all messages have been acked and the commit offset has been acknowledged by the
+  // Discard all outstanding acks and wait for any pending commit offset to be acknowledged by the
   // server. Throws an exception if the committer has shut down.
-  void waitUntilEmpty() throws CheckedApiException;
+  void waitUntilCommitted() throws CheckedApiException;
 }
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTracker.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTracker.java
@@ -26,6 +26,6 @@ interface AckSetTracker extends ApiService {
   Runnable track(SequencedMessage message) throws CheckedApiException;
 
   // Discard all outstanding acks and wait for any pending commit offset to be acknowledged by the
-  // server. Throws an exception if the committer has shut down.
+  // server. Throws an exception if the committer shut down due to a permanent error.
   void waitUntilCommitted() throws CheckedApiException;
 }
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
@@ -113,6 +113,7 @@ public class AckSetTrackerImpl extends TrivialProxyService implements AckSetTrac
     try (CloseableMonitor.Hold h = monitor.enter()) {
       receiptsCopy = ImmutableList.copyOf(receipts);
     }
+    // Clearing receipts here avoids deadlocks due to locks acquired in different order.
     receiptsCopy.forEach(Receipt::clear);
     try (CloseableMonitor.Hold h = monitor.enter()) {
       receipts.clear();
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
@@ -109,8 +109,8 @@ public class AckSetTrackerImpl extends TrivialProxyService implements AckSetTrac
       receipts.forEach(Receipt::clear);
       receipts.clear();
       acks.clear();
+      committer.waitUntilEmpty();
     }
-    committer.waitUntilEmpty();
   }
 
   private void onAck(Offset offset) {
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
@@ -28,7 +28,6 @@ import com.google.cloud.pubsublite.internal.CloseableMonitor;
 import com.google.cloud.pubsublite.internal.ExtractStatus;
 import com.google.cloud.pubsublite.internal.TrivialProxyService;
 import com.google.cloud.pubsublite.internal.wire.Committer;
-import com.google.common.util.concurrent.Monitor.Guard;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import java.util.ArrayDeque;
 import java.util.Deque;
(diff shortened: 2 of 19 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTracker.java
```diff
@@ -24,4 +24,8 @@
   // Track the given message. Returns a Runnable to ack this message if the message is a valid one
   // to add to the ack set. Must be called with strictly increasing offset messages.
   Runnable track(SequencedMessage message) throws CheckedApiException;
+
+  // Discard all outstanding acks and wait for any pending commit offset to be acknowledged by the
```
dpcollins-google, line 28: Instead, just make this extend Flushable like the internal publisher? https://docs.oracle.com/javase/8/docs/api/index.html?java/io/Flushable.html

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
```diff
@@ -36,13 +36,51 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class AckSetTrackerImpl extends TrivialProxyService implements AckSetTracker {
+  // Receipt represents an unacked message. It can be cleared, which will cause the ack to be
+  // ignored.
+  private static class Receipt implements Runnable {
```
dpcollins-google, line 41: Runnable is the one interface you never need to implement explicitly, you can just use method references if you need a runnable. Just remove the interface, rename the method as "onAck", and return `receipt::onAck`

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTracker.java
dpcollins-google: Actually ignore this.

tmdiep: Ack. I was going to say that I need to throw CheckedApiException.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
tmdiep: Done

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
```diff
@@ -36,13 +36,50 @@
...
 
 public class AckSetTrackerImpl extends TrivialProxyService implements AckSetTracker {
+  // Receipt represents an unacked message. It can be cleared, which will cause the ack to be
+  // ignored.
+  private static class Receipt {
+    private final CloseableMonitor m = new CloseableMonitor();
+    private final AtomicBoolean wasAcked = new AtomicBoolean(false);
+    final Offset offset;
+
+    @GuardedBy("m.monitor")
```
dpcollins-google, line 46: you violate this constraint on line 69 and 72. Instead, do the following:

1) hold the monitor for the full call of onAck
2) change the atomicBoolean for a normal boolean
3) change `waitUntilCommitted` to:
```
List<Receipt> receiptsCopy;
  receiptsCopy = ImmutableList.copyOf(receipts);
}
receiptsCopy.forEach(Receipt::clear);
      receipts.clear();
      acks.clear();
      committer.waitUntilEmpty();
    }
```

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/CommitterImpl.java
```diff
@@ -144,4 +151,13 @@ public void onClientResponse(SequencedCommitCursorResponse value) throws Checked
       return ApiFutures.immediateFailedFuture(e);
     }
   }
+
+  @Override
+  public void waitUntilEmpty() throws CheckedApiException {
+    try (CloseableMonitor.Hold h = monitor.enterWhenUninterruptibly(isEmptyOrShutdown)) {
+      if (shutdown) {
+        throw new CheckedApiException("Shutting down.", Code.UNAVAILABLE);
```
dpcollins-google, line 159: this will cause clean shutdown to become unclean. If we're trying to shut down cleanly, the queue of acks being flushed to the server should become empty over time like normal. If we're failed, you should fail here with the error it failed with.

What you should do is 1) change hadPermanentError to Optional<CheckedApiException> permanentError, 2) use the existing isEmptyOrError condition to check and test permanentError.isPresent when determining if you're going to throw.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/NextOffsetTracker.java
```diff
@@ -59,4 +59,9 @@ void onMessages(Collection<SequencedMessage> messages) throws CheckedApiExceptio
                 .setCursor(Cursor.newBuilder().setOffset(offset.value()))
                 .build());
   }
+
+  // Resets the offset tracker to its initial state.
```
dpcollins-google, line 63: remove. instead, just do `nextOffsetTracker = new NextOffsetTracker()`

dpcollins-google: Ah never mind. this prevents it from being final.

dpcollins-google:
> The next PR will fully handle admin seeks by setting InitialSubscribeRequest.initial_location. Omitted to keep the size of the PR down.

This leaves the client in an indeterminate state where it will drop a bunch of acks but not actually respect the seek on the reconnect. I don't think this is a good state to be in at head. Can you remove the changes to SubscriberImpl from this PR and move them to the next one, so at least if someone happens to (I don't think they will) pick up this intermediate state it will have the same behavior as the existing code (i.e. not reset?)

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
tmdiep: Done

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/CommitterImpl.java
tmdiep: Done

tmdiep:
> Can you remove the changes to SubscriberImpl from this PR and move them to the next one, so at least if someone happens to (I don't think they will) pick up this intermediate state it will have the same behavior as the existing code (i.e. not reset?)

Good point. Instead I just changed SinglePartitionSubscriber.onSubscriberReset to return false, which effectively becomes a no-op.

This will keep the size of the next PR down and avoids re-reviewing many files.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
```diff
@@ -57,23 +99,26 @@ public AckSetTrackerImpl(Committer committer) throws ApiException {
...
-        onAck(messageOffset);
-      }
-    };
+  @Override
+  public void waitUntilCommitted() throws CheckedApiException {
+    List<Receipt> receiptsCopy;
+    try (CloseableMonitor.Hold h = monitor.enter()) {
+      receiptsCopy = ImmutableList.copyOf(receipts);
+    }
+    receiptsCopy.forEach(Receipt::clear);
```
dpcollins-google, line 116: nit. add a comment that we're doing this to avoid lock order problems

dpcollins-google (approved):

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AckSetTrackerImpl.java
tmdiep: Done

PR #664, opened 2021-06-04
feat: Set initial location when connecting subscribe streams

tmdiep:
* Sets the `InitialSubscribeRequest.initial_location` field when reconnecting subscribe streams.
  * This speeds up reconnects by seeking to the next offset in the initial subscribe, rather than as a subsequent seek request.
  * Setting the initial location to `COMMITTED_CURSOR` also acknowledges the client's receipt of the `RESET` signal.
* Completes the handling of admin/out of band seeks.
* Removes the initial seek from all subscriber wrapper implementations. When dependent libraries (e.g. Spark, Kafka) are updated to a new release of this library, they will be updated to set the initial location for `SubscriberBuilder`.


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
```diff
@@ -230,25 +240,18 @@ public void triggerReinitialize(CheckedApiException streamError) {
...
+      connection.reinitialize(getInitialRequest());
       connection.modifyConnection(
           connectedSubscriber -> {
             checkArgument(monitor.monitor.isOccupiedByCurrentThread());
             checkArgument(connectedSubscriber.isPresent());
             if (inFlightSeek.isPresent()) {
               connectedSubscriber.get().seek(inFlightSeek.get().seekRequest);
             } else {
-              nextOffsetTracker
+              flowControlBatcher
```
dpcollins-google, line 251: I think this behavior is slightly incorrect. If there is an in flight seek, we still want to send the flow control restart request, since it may contain tokens sent after the seek was initiated.

dpcollins-google (approved):

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/SubscriberImpl.java
tmdiep: I realized that there was a subtle bug before - SubscriberImpl.seek() clears flow control tokens when the seek is initiated. Since the server hasn't received the seek request yet, the client may still receive messages, which will cause a permanent failure when TokenCounter.sub() detects underflow.

So we should actually clear flow control tokens upon seek response, and the caller should refill them after the seek future completes. I added a comment to the Subscriber interface to make this clear.

Unless we should have originally *not* cleared flow control tokens when there is a client seek? [This](https://github.com/googleapis/java-pubsublite/commit/bad81a55621bb2f42907cb261c3f32577c95af12#diff-92336fc6906a73e0298dbc24aa58b802a623261ffe8584a3fb4f364dd8f5697e) was the last patch related to this - but the tokens were already being reset upon seek initiation.

Fortunately, all current callers of seek call it after the initial subscribe, wait for the future, and then fill the flow control tokens. So this hasn't caused any issues.

PR #674, opened 2021-06-08
feat: Change restriction to OffsetByteRange to allow functioning with runnerv2.

### LINKED ISSUE #676: Crashes under runnerv2 due to project tracking semantics
Under dataflow RunnerV2, after trySplit returns successfully, the runner does not immediately check the progress accounted for by testing the RestrictionTracker. Instead, it destroys the RestrictionTracker and tries to construct a new one using the already claimed restriction instead. The runner will not do this if the SDF implements the @GetSize annotated method, in which case it will be used instead. As far as I can tell, this particular interplay is undocumented, but it should remain true for this public API.

Our I/O will crash when constructed with an already completed restriction to avoid constructing a restriction tracker in an unusable state. We should implement GetSize so that we don't need to allow for partially unusable RestrictionTrackers.

dpcollins-google:
Fixes #676

### DIFF
--- a/pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/OffsetByteRange.java
+++ b/pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/OffsetByteRange.java
@@ -16,16 +16,23 @@
 
 package com.google.cloud.pubsublite.beam;
 
-import com.google.cloud.pubsublite.proto.SequencedMessage;
-import java.io.Serializable;
-import org.apache.beam.sdk.transforms.DoFn.OutputReceiver;
-import org.apache.beam.sdk.transforms.splittabledofn.RestrictionTracker;
+import com.google.auto.value.AutoValue;
+import org.apache.beam.sdk.io.range.OffsetRange;
+import org.apache.beam.sdk.schemas.AutoValueSchema;
+import org.apache.beam.sdk.schemas.annotations.DefaultSchema;
 
-interface SubscriptionPartitionProcessorFactory extends Serializable {
-  long serialVersionUID = 765145146544654L;
+@AutoValue
+@DefaultSchema(AutoValueSchema.class)
+abstract class OffsetByteRange {
+  abstract OffsetRange getRange();
 
-  SubscriptionPartitionProcessor newProcessor(
-      SubscriptionPartition subscriptionPartition,
-      RestrictionTracker<OffsetByteRange, OffsetByteProgress> tracker,
-      OutputReceiver<SequencedMessage> receiver);
+  abstract long getByteCount();
+
+  static OffsetByteRange of(OffsetRange range, long byteCount) {
+    return new AutoValue_OffsetByteRange(range, byteCount);
+  }
+
+  static OffsetByteRange of(OffsetRange range) {
+    return of(range, 0);
+  }
 }
--- a/pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/OffsetByteRangeTracker.java
+++ b/pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/OffsetByteRangeTracker.java
@@ -26,8 +26,6 @@ import com.google.common.base.Stopwatch;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import org.apache.beam.sdk.io.range.OffsetRange;
-import org.apache.beam.sdk.transforms.splittabledofn.RestrictionTracker;
-import org.apache.beam.sdk.transforms.splittabledofn.RestrictionTracker.HasProgress;
 import org.apache.beam.sdk.transforms.splittabledofn.SplitResult;
 import org.joda.time.Duration;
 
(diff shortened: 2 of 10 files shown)


pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/OffsetByteRangeTracker.java
```diff
@@ -86,32 +87,32 @@ public boolean tryClaim(OffsetByteProgress position) {
...
+      boolean isRangeEmpty = range.getRange().getTo() == range.getRange().getFrom();
+      boolean isValidClosedRange = nextOffset() == range.getRange().getTo();
       checkState(
           isRangeEmpty || isValidClosedRange,
           "Violated class precondition: offset range improperly split. Please report a beam bug.");
       return false;
     }
     lastClaimed = toClaim;
-    byteCount += position.batchBytes();
+    range = OffsetByteRange.of(range.getRange(), range.getByteCount() + position.batchBytes());
```
boyuanzz, line 104: Why do you want to create a new `OffsetByteRange` every time you claim a position? You can change the `byteCount` directly, no?

pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/OffsetByteRange.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.google.cloud.pubsublite.beam;
+
+import com.google.auto.value.AutoValue;
+import org.apache.beam.sdk.io.range.OffsetRange;
+import org.apache.beam.sdk.schemas.AutoValueSchema;
+import org.apache.beam.sdk.schemas.annotations.DefaultSchema;
+
+@AutoValue
+@DefaultSchema(AutoValueSchema.class)
```
boyuanzz, line 25: Just FYI this change will introduce backward incompatibility for updating a Dataflow pipeline.

boyuanzz (commented):
Are you going to fix the coder issue with `setPartition()` in this PR or a separate PR?

dpcollins-google:
> Are you going to fix the coder issue with `setPartition()` in this PR or a separate PR?

It has already been removed since it does not provide an access pattern we want to expose.

boyuanzz:
> > Are you going to fix the coder issue with `setPartition()` in this PR or a separate PR?
> 
> It has already been removed since it does not provide an access pattern we want to expose.

Then you should remove this from Beam as well. Or if you decide to keep your own version, you should consider removing your code from Beam.

pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/OffsetByteRangeTracker.java
boyuanzz: Let me rephrase my statement:
You can keep `byteCount` as a local var to the tracker and increase it when `tryClaim`. The you port it back to restriction when trySplit() is called and tryClaim() return false.

pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/OffsetByteRangeTracker.java
```diff
@@ -123,29 +124,33 @@ private boolean receivedEnough() {
...
       return null;
     }
-    range = new OffsetRange(currentRestriction().getFrom(), nextOffset());
-    return SplitResult.of(this.range, new OffsetRange(nextOffset(), Long.MAX_VALUE));
+    range =
+        OffsetByteRange.of(
+            new OffsetRange(currentRestriction().getRange().getFrom(), nextOffset()),
+            range.getByteCount());
+    return SplitResult.of(
+        this.range, OffsetByteRange.of(new OffsetRange(nextOffset(), Long.MAX_VALUE), 0));
```
boyuanzz, line 147: By returning `OffsetByteRange.of(new OffsetRange(nextOffset(), Long.MAX_VALUE), 0)`, you are telling dataflow there is no backlog for this read. This will prevent dataflow brings more workers to work on this.

dpcollins-google:
> > > Are you going to fix the coder issue with `setPartition()` in this PR or a separate PR?
> > 
> > 
> > It has already been removed since it does not provide an access pattern we want to expose.
> 
> Then you should remove this from Beam as well. Or if you decide to keep your own version, you should consider removing your code from Beam.

Yes, I've already created https://github.com/apache/beam/pull/14976 to copy this and other changes over.

pubsublite-beam-io/src/main/java/com/google/cloud/pubsublite/beam/OffsetByteRangeTracker.java
dpcollins-google: I don't believe so? GetSize delegates to the underlying OffsetByteRangeTracker if the range is unbounded on the right (like this is) and so will call out to `this.backlogReader.computeMessageStats(range.getRange().getFrom())` to compute forward looking progress stats. For the existing range, it ships the byte size along with it which will be used since the already-finished range is closed. Is this analysis incorrect given how the runner interacts with these methods?

boyuanzz: That's correct.

boyuanzz:
Changes which are related to beam look good to me.

manuelmenzella-google (approved):

PR #692, opened 2021-06-11
cleanup: Simplify the internals of the java client library by removing data plane seek from the wire subscriber

dpcollins-google:
This feature is entirely unused.


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/ConnectedSubscriberImpl.java
```diff
@@ -122,23 +88,18 @@ protected void handleStreamResponse(SubscribeResponse response) throws CheckedAp
...
-      if (seekInFlight) {
-        log.atInfo().log(
-            "Dropping %s messages due to outstanding seek on stream with initial request %s.",
-            response.getMessagesCount(), initialRequest);
-        return;
-      }
-    }
     checkState(
         response.getMessagesCount() > 0,
         String.format(
```
tmdiep, line 105: Does `PullResponse` need to be updated, or deliberately used in the error message?

tmdiep (approved):

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/ConnectedSubscriberImpl.java
dpcollins-google: Done.

PR #693, opened 2021-06-11
cleanup: Modify pom files for java-pubsublite to enable errorprone analysis and general clean up

dpcollins-google:


google-cloud-pubsublite/pom.xml
```diff
@@ -195,10 +183,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>flatten-maven-plugin</artifactId>
-      </plugin>
```
anguillanneuf, line 201: Pinging our flatten maven plugin expert @stephaniewang526 here. Does flatten maven plugin conflict with errorprone?

stephaniewang526: No, it shouldn't. I would not recommend removing it.

dpcollins-google: This isn't removed, just moved to the plugins section on the parent POM

stephaniewang526: Okay as long as you're okay with generating flattened poms for parent/root poms as well; this is fine.

stephaniewang526 (approved):

anguillanneuf (approved):

PR #728, opened 2021-06-23
fix: Fail the partition watcher if its consumer throws an exception

palmere-google:

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PartitionCountWatcherImpl.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PartitionCountWatcherImpl.java
@@ -78,7 +78,12 @@ public class PartitionCountWatcherImpl extends AbstractApiService implements Par
     if (currentPartitionCount.isPresent() && currentPartitionCount.get().equals(partitionCount)) {
       return;
     }
-    partitionCountReceiver.accept(partitionCount);
+    try {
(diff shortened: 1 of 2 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PartitionCountWatcherImpl.java
```diff
@@ -78,7 +78,12 @@ private void pollTopicConfig() {
     if (currentPartitionCount.isPresent() && currentPartitionCount.get().equals(partitionCount)) {
       return;
     }
-    partitionCountReceiver.accept(partitionCount);
+    try {
+      partitionCountReceiver.accept(partitionCount);
+    } catch (Throwable t) {
+      notifyFailed(ExtractStatus.toCanonical(t));
+      stop();
```
dpcollins-google, line 85: should notifyFailed eventually trigger stop() to be called? if so, remove this call.

palmere-google: That's a good question, from the docs it seems like it doesn't

https://guava.dev/releases/23.0/api/docs/com/google/common/util/concurrent/AbstractService.html#notifyFailed-java.lang.Throwable-

"Invoke this method to transition the service to the Service.State.FAILED. The service will not be stopped if it is running. Invoke this method when a service has failed critically or otherwise cannot be started nor stopped."

dpcollins-google (approved):

PR #740, opened 2021-07-07
feat: Support seek subscription in AdminClient

tmdiep:
Seek subscription performs an out-of-band seek for a subscription to a specified target, which may be a backlog location, publish timestamp or event timestamp.

Note: feature is currently pre-release.


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/AdminClientImpl.java
```diff
@@ -188,6 +194,29 @@ public CloudRegion region() {
...
+  @Override
+  public OperationFuture<SeekSubscriptionResponse, OperationMetadata> seekSubscription(
+      SubscriptionPath path, SeekTarget target) {
+    SeekSubscriptionRequest.Builder request =
+        SeekSubscriptionRequest.newBuilder().setName(path.toString());
+    switch (target.getKind()) {
+      case BACKLOG_LOCATION:
+        if (target.backlogLocation() == BacklogLocation.END) {
+          request.setNamedTarget(NamedTarget.HEAD);
+        } else if (target.backlogLocation() == BacklogLocation.BEGINNING) {
```
dpcollins-google, line 206: make this a nested switch instead to ensure full coverage (the compiler will fail if you don't handle all cases)

dpcollins-google (approved):

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/AdminClientImpl.java
tmdiep: Done. Thanks Dan.

PR #744, opened 2021-07-09
fix: Change all DirectExecutors to use a SystemExecutor and fix service shutdowns

dpcollins-google:
Also deflake the world so all tests run 100 times in a row successfully even with the new executors, and centralize alarm creation to fix polling test bugs due to scheduleAtFixedRate and test flakiness.

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PartitionCountWatcherImpl.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PartitionCountWatcherImpl.java
@@ -15,102 +15,87 @@
  */
 package com.google.cloud.pubsublite.internal.wire;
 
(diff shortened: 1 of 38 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/PartitionCountWatcherImpl.java
```diff
@@ -15,102 +15,87 @@
...
-  private void stop() {
-    partitionCountPoll.cancel(true);
-    adminClient.close();
   }
 
   @Override
   protected void doStart() {
-    partitionCountPoll =
-        SystemExecutors.getAlarmExecutor()
-            .scheduleAtFixedRate(
```
dpcollins-google, line 103: This is one footgun that means there can be a data race if one of the poll requests locks up.

manuelmenzella-google (approved):

PR #799, opened 2021-08-04
chore: remove internal packages from javadoc

eaball35:
Config for both googleapis.dev and cloud-rad.
Excluding packages:
- com.google.cloud.pubsublite.cloudpubsub.internal
- com.google.cloud.pubsublite.internal
- com.google.cloud.pubsublite.internal.testing
- com.google.cloud.pubsublite.internal.wire

I tested both kokoro jobs doc generation scripts up to actually uploading docs and both were successful. 

Can test locally with:
- googleapis.dev: `mvn site` -> output in target/site/apidocs
- cloud.google.com: `mvn site -P docFX` -> output in target/docfx-yml

### DIFF
--- a/.kokoro/release/publish_javadoc.sh
+++ b/.kokoro/release/publish_javadoc.sh
@@ -48,6 +48,6 @@ python3 -m docuploader create-metadata \
   --language java
 
 # upload docs
-#python3 -m docuploader upload . \
-#  --credentials ${CREDENTIALS} \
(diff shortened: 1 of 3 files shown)


anguillanneuf:
Hm, it looks like more packages were excluded than intended:

Screen Shot 2021-08-05 at 9 48 14 AM
Screen Shot 2021-08-05 at 10 30 31 AM

anguillanneuf (approved):

PR #860, opened 2021-09-08
fix: Ensure notifyFailed is called and all services are stopped when a permanent error happens

dpcollins-google:

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/ProxyService.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/ProxyService.java
@@ -23,8 +23,10 @@ import static com.google.cloud.pubsublite.internal.UncheckedApiPreconditions.che
 import com.google.api.core.AbstractApiService;
 import com.google.api.core.ApiService;
 import com.google.api.gax.rpc.ApiException;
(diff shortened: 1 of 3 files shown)


hannahrogers-google (approved):

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/ProxyService.java
```diff
@@ -67,8 +70,10 @@ protected final void addServices(ApiService... services) throws ApiException {
...
   protected final void onPermanentError(CheckedApiException error) {
     if (stoppedOrFailed.getAndSet(true)) return;
-    for (ApiService service : services) {
-      service.stopAsync();
+    try {
+      ApiServiceUtils.stopAsync(services);
+    } catch (Throwable t) {
+      LOGGER.atFine().withCause(t).log("Exception in underlying service shutdown.");
     }
     handlePermanentError(error);
```
gpoulin-hopper, line 78: We should also likely handle exception from `handlePermanentError(error)` that was the line that actually caused issue for us.

dpcollins-google: Good point. Fixed.

PR #929, opened 2021-10-25
samples: regional topics samples

anguillanneuf:
DO NOT SUBMIT until service goes live (early 2022).

All tests pass using a whitelisted project for now.

Also see (Python) https://github.com/googleapis/python-pubsublite/pull/266


samples/snippets/src/main/java/pubsublite/ListTopicsExample.java
```diff
@@ -46,6 +46,7 @@ public static void listTopicsExample(String cloudRegion, char zoneId, long proje
     LocationPath locationPath =
         LocationPath.newBuilder()
             .setProject(ProjectNumber.of(projectNumber))
+            // To list topic in a region, set location to a cloud region instead.
```
hongalex, line 49: nit: s/topic/topics

anguillanneuf: Good catch. Thanks!

PR #946, opened 2021-11-10
deps: Correct BOM deps

dpcollins-google:


google-cloud-pubsublite-bom/pom.xml
```diff
@@ -7,8 +7,9 @@
   <packaging>pom</packaging>
   <parent>
     <groupId>com.google.cloud</groupId>
-    <artifactId>google-cloud-shared-config</artifactId>
-    <version>1.2.1</version>
+    <artifactId>google-cloud-pubsublite-parent</artifactId>
+    <version>1.3.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-pubsublite:current} -->
+    <relativePath>../pom.xml</relativePath>
```
chingor13, line 12: We do want to keep this parent to be directly the shared-config. Another example is https://github.com/googleapis/java-asset/blob/main/google-cloud-asset-bom/pom.xml

chingor13 (dismissed):

google-cloud-pubsublite-bom/pom.xml
dpcollins-google: Fixed.

manuelmenzella-google (approved):

PR #988, opened 2021-12-14
fix: Share the transport channel provider for publishers and subscribers in the same region

dpcollins-google:
This prevents GRPC lock ups that can happen when creating new channels all the time for publishers with many partitions.

Use 100 channels in the pool to avoid GRPC limits on streams per channel.


dpcollins-google:
Closing, this doesn't actually deduplicate the channels.

PR #1028, opened 2022-01-20
samples: add reservation samples

anguillanneuf:
Towards b/213558019
Adding Java samples for the following region tags:

- `pubsublite_create_reservation`
- `pubsublite_get_reservation`
- `pubsublite_update_reservation`
- `pubsublite_list_reservations`
- `pubsublite_delete_reservation`

Other languages to be added: Python, Go


samples/snippets/src/main/java/pubsublite/CreateReservationExample.java
```diff
@@ -0,0 +1,71 @@
...
+            .setThroughputCapacity(throughputCapacity)
+            .build();
+
+    AdminClientSettings adminClientSettings =
+        AdminClientSettings.newBuilder().setRegion(CloudRegion.of(cloudRegion)).build();
+
+    try (AdminClient adminClient = AdminClient.create(adminClientSettings)) {
+      Reservation response = adminClient.createReservation(reservation).get();
+      System.out.println(response.getAllFields() + " created successfully.");
+    } catch (AlreadyExistsException e) {
```
dpcollins-google, line 66: This wouldn't ever happen- it would be an ExecutionException I think?

anguillanneuf: You are right, it shows up as `Exception in thread "main" java.util.concurrent.ExecutionException: com.google.api.gax.rpc.AlreadyExistsException: io.grpc.StatusRuntimeException: ALREADY_EXISTS:` but should indeed be `AlreadyExistsException`? 

I can remove the catch now, but propose we surface the gax.rpc exceptions. Wdyt?

dpcollins-google: Then pubsub is likely also wrong. You need to catch the ExecutionException and look at the "cause()" member

dpcollins-google (approved):

PR #1286, opened 2022-12-08
feat: unload partitions when not publishing for better performance

dpcollins-google:


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/UnloadingPublisher.java
```diff
@@ -0,0 +1,117 @@
...
+          .get()
+          .addListener(
+              new Listener() {
+                @Override
+                public void failed(State from, Throwable failure) {
+                  notifyFailed(failure);
+                }
+              },
+              SystemExecutors.getFuturesExecutor());
+      publisher.get().startAsync().awaitRunning();
```
tmdiep, line 91: Is it ok to block the publish() call here, while the stream is connecting?

In Go, I allowed the wire publisher to accept messages while it is starting up. The messages just get queued until the stream is connected.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/UnloadingPublisher.java
```diff
@@ -0,0 +1,117 @@
...
+  @Override
+  public synchronized void flush() throws IOException {
+    if (publisher.isPresent()) {
+      publisher.get().flush();
+    }
+  }
+
+  private synchronized void onUnloadAlarm() {
+    if (publisher.isPresent() && !sawPublish) {
+      publisher.get().stopAsync().awaitTerminated();
```
tmdiep, line 112: Similar to the above, waiting here may block publish() due to the lock held.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/UnloadingPublisher.java
dpcollins-google: I think this issue already exists- so I'm okay with blocking here. We should as a refactor fix this eventually.

PR #1334, opened 2023-02-20
fix: Change proxy service to call child start methods on executors

dpcollins-google:
Also change service clients to use an isolated executor.
Also change channel pools to not use a fixed pool size, and use an auto-scaling channel pool.


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/ProxyService.java
```diff
@@ -57,6 +57,14 @@ protected final <T extends ApiService> void addServices(Collection<T> services)
     checkState(state() == State.NEW);
     for (ApiService service : services) {
       checkArgument(service.state() == State.NEW, "All services must not be started.");
+      service.addListener(
+          new Listener() {
```
tmdiep, line 61: Nit: use the same listener instance for all the dependencies?

tmdiep (approved):

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/ProxyService.java
dpcollins-google: Done.

PR #1335, opened 2023-02-20
feat: Convert internal interfaces to use protos

dpcollins-google:


tmdiep (approved):

jhsenjaliya:
This change breaks the compatibility with "com.google.cloud:pubsublite-spark-sql-streaming:1.0.0" which uses non-proto SequencedMessage

PR #1360, opened 2023-03-14
feat: enable compression for publishing

dpcollins-google:
This is enabled by default and should save on network usage for clients

Internal reference b/269065916

### DIFF
--- a/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/PublisherSettings.java
+++ b/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/PublisherSettings.java
@@ -22,7 +22,7 @@ import static com.google.cloud.pubsublite.internal.wire.ServiceClients.getCallCo
 
 import com.google.api.gax.batching.BatchingSettings;
 import com.google.api.gax.core.CredentialsProvider;
-import com.google.api.gax.rpc.ApiCallContext;
+import com.google.api.gax.grpc.GrpcCallContext;
 import com.google.api.gax.rpc.ApiException;
 import com.google.api.gax.rpc.StatusCode.Code;
 import com.google.auto.value.AutoValue;
@@ -50,6 +50,7 @@ import com.google.cloud.pubsublite.v1.PublisherServiceSettings;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.ByteString;
 import com.google.pubsub.v1.PubsubMessage;
+import io.grpc.CallOptions;
 import java.util.Optional;
 import org.threeten.bp.Duration;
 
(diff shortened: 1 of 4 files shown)


google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/PublisherSettings.java
```diff
@@ -87,6 +88,9 @@ public abstract class PublisherSettings {
    */
   abstract boolean enableIdempotence();
 
+  /** Whether to enable request compression. Default true. */
```
davidtorres, line 91: This is a getter, shouldn't this read as "Whether request compression is enabled."

I'd also fix the wording for enableIdempotence()

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/PublisherSettings.java
```diff
@@ -142,6 +147,12 @@ public abstract Builder setMessageTransformer(
      */
     public abstract Builder setEnableIdempotence(boolean enableIdempotence);
 
+    /**
+     * Whether to enable publish idempotence, where the server will ensure that unique messages
```
davidtorres, line 151: Fix this javadoc.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/PublisherSettings.java
```diff
@@ -190,10 +201,14 @@ public com.google.cloud.pubsublite.internal.Publisher<MessageMetadata> newPublis
                 .setPartition(partition)
                 .setStreamFactory(
                     responseStream -> {
-                      ApiCallContext context =
+                      GrpcCallContext context =
                           getCallContext(
                               PubsubContext.of(framework()),
                               RoutingMetadata.of(topicPath(), partition));
+                      if (enableCompression()) {
+                        context =
```
davidtorres, line 209: In CPS we don't just blindly enable compression and instead use a threshold based on the payload size as load tests showed that enabling it for small payloads was not gaining anything but actually hurting the CPU utilization.

davidtorres (changes requested):

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/PublisherSettings.java
dpcollins-google: We can't do that since our RPCs are actually streams.

google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/PublisherSettings.java
dpcollins-google: Done.

davidtorres (approved):

PR #1728, opened 2024-09-26
chore: use composite action

JoeWang1127:
In this PR:

- Use composite action defined in sdk-platform-java to reduce code duplication


.github/workflows/hermetic_library_generation.yaml
```diff
@@ -17,24 +17,29 @@ name: Hermetic library generation upon generation config change through pull req
...
-      run: |
-        set -x
-        [ -z "$(git config user.email)" ] && git config --global user.email "cloud-java-bot@google.com"
-        [ -z "$(git config user.name)" ] && git config --global user.name "cloud-java-bot"
-        bash .github/scripts/hermetic_library_generation.sh \
-          --target_branch ${{ github.base_ref }} \
-          --current_branch ${{ github.head_ref }}
-      env:
-        GH_TOKEN: ${{ secrets.CLOUD_JAVA_BOT_TOKEN }}
+    - uses: googleapis/sdk-platform-java/.github/scripts@v2.46.0
```
blakeli0, line 40: Would there be another PR to group this with sdk-platform-java upgrade? Or maybe we can do it in this PR?

blakeli0: Never mind, I remember it would be part of the renovate bot template change.

blakeli0 (approved):

